PR #2, opened 2020-03-19
add redux and propose a structure of redux store and model

mat10tng:
* Added a redux store
    * Added a redux dev tools extensions.
* Propose way of using redux thunk instead of redux-bees
* Propose a structure way of dealing with models ( domain base architecture)  
    */js/store/models


utunga:
@Kooshaba could you take a look at this? Looks good to me

Issue #6, opened 2020-03-21
Sign-up screen [In Progress by Terrence]


mat10tng:
![image](https://user-images.githubusercontent.com/3273671/77228454-e847d480-6b87-11ea-99eb-f759df125b64.png)
Link to Figma design: [Link](https://www.figma.com/file/v5HdxDsvO2NzMuZU6CArzF/Corona-Donor-Designs-in-progress?node-id=170%3A255)
Please refer to figma design as it has changed! The picture uploaded are only used as a point of ref.

CSumm:
@mat10tng are you doing this issue, or is it possible for me to take up the job?

mat10tng:
Hi, please check with @Terrence  in slack channel, He said he was ready to take it up. But I have no clue. I will try to get some more issues going

PR #17, opened 2020-03-22
Add Signup Page

thedma04:
This PR addresses issue #6


mat10tng:
Yeah, its look good. I am actually work on the same type of components, and so next time we should use FormInput instead of directly use TextInput

              <FormInput inputObj={{
                      labelPhrase: 'Short Description'
              }}/>

PR #27, opened 2020-03-23
Add Input component with test and stories with knob

ruchijn:
Input - text:
input_text

Input-textarea:
input_textarea

### DIFF
--- a/src/app/component/Button/utils.js
+++ b/src/app/component/Button/utils.js
@@ -1,21 +1,3 @@
-export const colors = {
-  primary: {
-    background: '#140D60'
-  },
-  secondary: {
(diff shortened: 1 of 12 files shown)


src/app/component/Button/utils.js
```diff
@@ -1,21 +1,3 @@
...
-
-  tertiary: {
-    background: '#F3D3EE'
-  },
-  disabled: {
-    border: '#e5e5e5',
-    color: 'rgba(0, 0, 0, 0.2)'
-  }
-};
-
```
ruchijn: Moved colors out to constants file so that they can be used in other componnets.

PR #33, opened 2020-03-23
Feat/image uploader

### LINKED ISSUE #23: Photo upload component story // In Progress - Terrence
image

thedma04:
Fixes #23

No image
image

When an image is taken or uploaded
image


src/app/component/ImageUpload/ImageUpload.test.js
```diff
@@ -0,0 +1,34 @@
...
+import { ImageUpload } from '.';
+
+describe('ImageUpload Component', () => {
+  afterEach(cleanup);
+
+  global.URL.createObjectURL = jest.fn(); // Mock
+
+  it('renders camera icon and two buttons', () => {
+    const { getByText, getByTestId } = render(<ImageUpload />);
+    expect(getByTestId('cameraIcon')).toBeTruthy();
```
aamin21: Minor: I think explicit assertion i.e. toBe(true) is better than Truthy tests as the latter sometimes give false positives i.e. empty objects evaluate to true

src/app/component/ImageUpload/index.js
```diff
@@ -0,0 +1,82 @@
+import React, { useRef, useState } from 'react';
```
aamin21: This one is for the team: while using index.js inside a named dir is standard, I believe it adds more issues. For example, in my IDE, all the files would be named index.js. I would need to hover on the filename to get the dir name. This is specially annoying when you need to search up a file name, but can't since every file is named index.js 

But again, this is for the whole team to decide.

src/app/component/ImageUpload/index.js
```diff
@@ -0,0 +1,82 @@
...
+import ImagePreview from './ImagePreview';
+import Button from '../Button/Button';
+
+export function ImageUpload({ styles, getFile = () => null }) {
+  const fileUpload = useRef();
+  const [file, setFile] = useState('');
+  const [preview, setPreview] = useState(null);
+
+  function handleFileChange({ target }) {
+    setFile(target.files[0]);
```
aamin21: Should we do a sanity check before setting and getting the file i.e. target.files.length && setFile()

src/app/component/ImageUpload/index.js
mat10tng: Yeah, agree. I think we will likely keep index.js file only for import and export in the future

src/app/component/ImageUpload/index.js
snowcoding: Maybe we could set a default value for the target, which could simplify the sanity check?

src/app/component/ImageUpload/index.js
```diff
@@ -0,0 +1,82 @@
...
+        ref={fileUpload}
+        type="file"
+        accept="image/x-png,image/gif,image/jpeg capture=camera"
+        hidden
+      />
+    </Container>
+  );
+}
+
+const Container = styled.div`
```
snowcoding: Based on quick chat from @mat10tng, we might want to move the styled components into ImageUpload.style.js, if I understood him correctly

mat10tng: yes, I have had some discussion with Terrence.

src/app/component/ImageUpload/ImageUpload.test.js
thedma04: Using toBe(true) will not be the best option here since toBe will compare it to a string instead checking whether it's available. 
image

The best matcher to use is `toBeInTheDocument`

PR #37, opened 2020-03-23
chore(dev): remove yarn.lock- stick with npm- and integrate prettier …

considine:
Removes yarn.lock so we're all on the same page w.r.t package managers
Also integrates [Prettier Pre-Commit Hook](https://prettier.io/docs/en/precommit.html) so committed files are linted in a standardized way


snowcoding:
Hey Jack - thanks for getting this going. It's definitely needed.

The [CRA docs](https://create-react-app.dev/docs/setting-up-your-editor#formatting-code-automatically) recommend something like this:

```
"lint-staged": {
    "src/**/*.{js,jsx,json,css,scss,md}": [
      "prettier --write"
    ]
  },
```

What do you think about this? It reduces scope to just the src file.

Secondly, were you able to test this locally at all?

Lastly, I think if you ran this:
```
./node_modules/.bin/prettier --write "src/**/*.{js,jsx,json,css,scss,md}"
```
it would update all files and then we could have a good starting point. Might need to run tests again as well before pushing.

Looking forward to hearing your thoughts, Cheers!

mat10tng:
I agree with snowcoding about the test. Can we add pre hook of npm tests as well ?

considine:
Sounds good @snowcoding @mat10tng ,

Just updated the hook, tested it manually, linted everything (like @snowcoding said) and ran npm tests- all still passing

snowcoding:
@considine, nice work

@mat10tng, LGTM, I'm not sure of the merging policies, so I will leave that to you.

considine:
@snowcoding thanks! don't have merge access @utunga could you please merge this

utunga:
👍  it's a beautiful thing...

PR #39, opened 2020-03-23
Refactor/routing

nickdurbin:
## Description

Extracted the routing out to its own component. This will allow for easy maintainability and scalability. Currently, I created a PrivateRoute, but this will probably need edited depending on how authorization and authentication are done. It is currently setup for JWT's. 

Also, a NotFound Route/Component to catch any urls hit that we do not have... This can get some styling with an error printed.

## Changes

Refactor 

## Testing 

None


src/app/router/PrivateRoute.js
```diff
@@ -0,0 +1,25 @@
...
+// Current setup is for tokens wihtout refresh tokens
+
+const PrivateRoute = props => {
+  const { component: Component, ...rest } = props;
+
+  return (
+    <Route
+      {...rest}
+      render={renderProps => {
+        if (localStorage.getItem("token")) {
```
azmainamin: Tests would be awesome for this component!

src/app/router/Router.js
```diff
@@ -0,0 +1,30 @@
...
+import routes from './routes';
+import { Route, BrowserRouter, Switch } from 'react-router-dom';
+import PrivateRoute from './PrivateRoute';
+import NotFound from '../page/NotFound';
+
+const Router = () => {
+  return (
+    <BrowserRouter>
+      <Switch>
+        {routes.map(({ key, path, component, requiresAuth, exact }) =>
```
azmainamin: Again, testing the conditional logic seems warranted.

src/app/page/NotFound/NotFound.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+const NotFound = () => {
+  return (
+    <div>
```
ruchijn: Also, we can simply do 

```
const NotFound = () => (
    <div>
      ...
    </div>
)
```

nickdurbin:
Going to work on the testing for the logic components. Will make a new PR when testing is complete! 

Thank you!

PR #44, opened 2020-03-24
Google map implementation

CSumm:
Resubmission of Google Maps implementation that includes:
-Google Map as minimap React component that uses google-map-react
-MiniMap stories.js file that can be used with storybook

Question along with this PR is: are component tests mandatory? I can try to add one, but I have not done it before and not sure which library we are using for it.


src/app/component/Minimap/MiniMap.stories.js
```diff
@@ -0,0 +1,10 @@
+import React from "react";
+import { storiesOf } from "@storybook/react";
+import MiniMap from './MiniMap';
+
+storiesOf("Components | Button", module)
```
mat10tng: We are now trying to write stories the same as Input Component, with Knobs decorator

src/app/component/Minimap/MiniMap.jsx
```diff
@@ -0,0 +1,80 @@
...
+// Module imports
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+import GoogleMapReact from "google-map-react";
+import styled from 'styled-components';
+/*** [end of imports] ***/
+
+
+//size variations for maps
+const SmallMap = styled.div`
```
mat10tng: style  should go into MiniMap.style.js

mat10tng:
Hi! there is some conflicts of package.json, once you solve that, I will merge in the pr

src/app/component/Minimap/MiniMap.stories.js
CSumm: I added knobs decorator, however I noticed when changing inside of storybook, the props do not resize right away. It only resize if you write in stories.js code

src/app/component/Minimap/MiniMap.jsx
CSumm: Moved to new file called MiniMap.style.js

CSumm:
I feel I might be missing something from the test however I am not entirely sure. A review would help. I'll also try and study Jest a bit more.

mat10tng:
we are using other maps now?

CSumm:
We'll use other map, but will still use Google's geocoding api to convert addresses to latitude and longitude. Since its a later priority, I can hold off with this PR for now

Issue #60, opened 2020-03-25
Input Component enhancement


mat10tng:
add a new prop :
* required: boolean
when true, and field is empty, it should tell the user that the field is required

Raf-Batista:
I can try this one if no one is on it.

thirdeyeclub:
> 
> 
> I can try this one if no one is on it.
I was gonna PR this one in a few minutes

thirdeyeclub:
> 
> 
> I can try this one if no one is on it.

Go ahead I couldn't make it work with the way it's currently written

mat10tng:
You are welcome to make any changes to make it work!

Raf-Batista:
This is how I currently have it implemented:

I used state which is updated with onChange

Every input that doesn't have a value when the form is submitted will have the placeholder set to "REQUIRED"

I've never made a PR so I have to do some reading on that. Once I figure that out I will take care of it. I will try to find a better solution to this as well.

mat10tng:
https://github.com/factn/mutualaidworld_frontend/blob/master/CONTRIBUTING.md

utunga:
@Raf-Batista super hoping that this process is not actually too hard for folks that havent made a PR before.. so
(a) if you figure it out let me know what bits were missing from the instructions 
(b) if you get stuck just let me/us know (that said im heading to bed just now,.. but yeah)

Either way, keen to hear how you get on..

utunga:
If you already made your changes locally, probably the easiest way to get back on track is to follow the instructions to create a fork, create a second local copy and just copy your changes in to that directory.

Raf-Batista:
Thanks for the help. I submitted the PR following a guide. I will post the link to it in the frontend channel for anyone who's new to this like me. I appreciate the help and welcome any sort of feedback, advice, and critiques. I want to keep improving and help with this project as much as I can. Cheers!

PR #71, opened 2020-03-26
Input Component enhancement #60

Raf-Batista:
Inputs can now use a placeholder and the Signup component now has state which is updated with the value of the input. The inputs in Signup.js have a placeholder set to the state of the same name as the input (which is initially an empty string).  When a user submits the form, we check the state for any empty values. For every empty value, we update the state with a  value of 'REQUIRED'.  Since the input has a placeholder set to the state, each empty input will have a  placeholder with 'REQUIRED'.

### DIFF
--- a/src/app/page/Signup/Signup.js
+++ b/src/app/page/Signup/Signup.js
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useState } from "react";
 import { Row, Col } from "react-flexbox-grid";
 
 import { Page } from "../../layout";
@@ -7,14 +7,37 @@ import Button from "../../component/Button";
 import { FormWrapper, HeaderText, DescriptionText, PaddedDiv } from "./Signup.style";
 
 const SignupPage = () => {
+  const [form, setForm] = useState({
+    fullName: "",
+    email: "",
+    phoneNumber: "",
+    zipCode: "",
+    errors: "",
+  });
+
   const handleLoginCTAClick = (e) => {
     e.preventDefault();
 
     window.location = "/login";
   };
 
-  const handleFormSubmit = () => {
+  const handleFormSubmit = (e) => {
+    e.preventDefault();
     console.log("Submit form and redirect user");
+    let newState = {};
+
+    for (let field in form) {
+      if (!form.field) newState[field] = "REQUIRED";
+    }
+
+    setForm(newState);
+  };
+
+  const handleChange = (e) => {
+    setForm({
+      ...form,
+      [e.target.name]: e.target.value,
+    });
   };
 
   return (
@@ -22,16 +45,32 @@ const SignupPage = () => {
       <FormWrapper>
         <HeaderText>Create account</HeaderText>
         <DescriptionText>Create an account to post your request</DescriptionText>
-        <form onSubmit={() => handleFormSubmit()}>
+        <form onSubmit={(e) => handleFormSubmit(e)}>
           <Row>
             <Col xs={12}>
               <PaddedDiv>
-                <Input inputType="text" dataId="fullName" inputName="fullName" label="FULL NAME" />
+                <Input
+                  inputType="text"
+                  dataId="fullName"
+                  inputName="fullName"
+                  label="FULL NAME"
+                  placeholder={form["fullName"]}
+                  value={form["fullName"]}
+                  onChange={(e) => handleChange(e)}
+                />
               </PaddedDiv>
             </Col>
             <Col xs={12}>
               <PaddedDiv>
-                <Input inputType="text" dataId="email" inputName="email" label="EMAIL" />
+                <Input
+                  inputType="text"
+                  dataId="email"
+                  inputName="email"
+                  label="EMAIL"
+                  placeholder={form["email"]}
+                  value={form["email"]}
+                  onChange={(e) => handleChange(e)}
+                />
               </PaddedDiv>
             </Col>
             <Col xs={12}>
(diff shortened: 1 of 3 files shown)


src/app/page/Signup/Signup.js
```diff
@@ -7,31 +7,70 @@ import Button from "../../component/Button";
...
+    email: "",
+    phoneNumber: "",
+    zipCode: "",
+    errors: "",
+  });
+
   const handleLoginCTAClick = (e) => {
     e.preventDefault();
 
     window.location = "/login";
```
thedma04: not part of your changes though but can with use the `history` props given by react-router as window.location will reload the whole page

src/app/page/Signup/Signup.js
```diff
@@ -7,31 +7,70 @@ import Button from "../../component/Button";
 import { FormWrapper, HeaderText, DescriptionText, PaddedDiv } from "./Signup.style";
 
 const SignupPage = () => {
+  const [form, setForm] = useState({
```
thedma04: There is a useForm hook inside the hooks folder that basically do the same thing
Can we use that for reusability? Thanks

src/app/page/Signup/Signup.js
```diff
@@ -7,31 +7,70 @@ import Button from "../../component/Button";
...
               <PaddedDiv>
-                <Input inputType="text" dataId="email" inputName="email" label="EMAIL" />
+                <Input
+                  inputType="text"
+                  dataId="email"
+                  inputName="email"
+                  label="EMAIL"
+                  placeholder={form["email"]}
+                  value={form["email"]}
+                  onChange={(e) => handleChange(e)}
```
mat10tng: Since onChange function take in the same variable as the input function you can also do
                  onChange={(e) => handleChange(e)}
To 
                  onChange={handleChange}

src/app/page/Signup/Signup.js
```diff
@@ -7,31 +7,70 @@ import Button from "../../component/Button";
...
-        <form onSubmit={() => handleFormSubmit()}>
+        <form onSubmit={(e) => handleFormSubmit(e)}>
           <Row>
             <Col xs={12}>
               <PaddedDiv>
-                <Input inputType="text" dataId="fullName" inputName="fullName" label="FULL NAME" />
+                <Input
+                  inputType="text"
+                  dataId="fullName"
+                  inputName="fullName"
```
mat10tng: input have alot of redundancy props
inputName and dataId is usually the same
since 
    value={form["fullName"]}
and you have the inputName = "fullName"

Maybe a better approach would be send in the entire **form** values and pick out the correct value with dataId/inputName variable

src/app/page/Signup/Signup.js
Raf-Batista: That's a much better approach. I'll get on it. Thanks

src/app/page/Signup/Signup.js
Raf-Batista: Can you point me to an article or somewhere I can read more about this? I never used hooks so  I don't quite understand your approach. Sorry.

src/app/page/Signup/Signup.js
Raf-Batista: I didn't know that. I'll take care of it. Thanks

src/app/page/Signup/Signup.js
Raf-Batista: I think I understand what you mean. I'll take care of the other changes first. After I take care of the changes what do I have to do? Do I have to submit another pull request? 

Sorry for the noobish question

src/app/page/Signup/Signup.js
Raf-Batista: I got it working with props. 

    props.history.push('/login');

I used withRouter but I think I can pass in the history prop from the parent. What do you suggest?

src/app/page/Signup/Signup.js
thedma04: Sure, there is a hook directory which contains useForm hook
It exposes the handleChange, setValues and values
You can use that instead of creating your own hook

Raf-Batista: Man, that is awesome. I had no idea you can do something like that. Explains why I completely missed the hooks directory. Thanks for pointing that out.

src/app/page/Signup/Signup.js
thedma04: I think you don't need withRouter, history is already passed from the Router component
Remove withRouter and console.log(prop.history). You will see it available

Raf-Batista: I tried console.log(props.history) and got undefined. I looked at App.tsx and noticed this 

  <Route path="/signup">
      <SignupPage />
  </Route>

If I change that to 

  <Route path="/signup" component={Signup}  />

It should pass in the history props right?

src/app/page/Signup/Signup.js
mat10tng: Yeah, just do another pr, and we will review it

src/app/page/Signup/Signup.js
thedma04: Yeah

src/app/page/Signup/Signup.js
Raf-Batista: I will submit another pr since I made changes that were suggested, I just haven't found a good way to take care of this one. I tried to send the values and pick the value with dataId/inputName but had trouble.

mat10tng:
We swich over to material-ui buttons

PR #72, opened 2020-03-26
Firebase phone signin

azmainamin:
feat: redesigned the login page to be consistent with the rest of the app. Created a phone authentication form. Authenticate using phone number


pedro-surf:
When I enter the code I received, I got redirected to a broken route.

PR #74, opened 2020-03-26
Consolidate signin flow

azmainamin:
feat: Remove Login/Sign Up header
feat: consolidate both actions using just Sign In using FB or Phone.
feat: some cosmetic changes.

image


mat10tng:
I think we must make it clear that you can sign in without having an account such as have a text that say sign in with phone number. Otherwise it might create confusion because it is different from the normal signup - signin flow

azmainamin:
> 
> 
> I think we must make it clear that you can sign in without having an account such as have a text that say sign in with phone number. Otherwise it might create confusion because it is different from the normal signup - signin flow

@mat10tng Agreed. Where should the text go? On the homepage or login page?

azmainamin:
> 
> 
> > I think we must make it clear that you can sign in without having an account such as have a text that say sign in with phone number. Otherwise it might create confusion because it is different from the normal signup - signin flow
> 
> @mat10tng Agreed. Where should the text go? On the homepage or login page?



> 
> 
> > I think we must make it clear that you can sign in without having an account such as have a text that say sign in with phone number. Otherwise it might create confusion because it is different from the normal signup - signin flow
> 
> @mat10tng Agreed. Where should the text go? On the homepage or login page?



> 
> 
> I think we must make it clear that you can sign in without having an account such as have a text that say sign in with phone number. Otherwise it might create confusion because it is different from the normal signup - signin flow

Right now, when you click on Sign In button in Home page, it takes you to the Sign In page:
image

this is already implemented. Let me know if you want me to change anything.

mat10tng:
Yeah, I am talking more from an ux perspective. When an user  come to the homepage and want to sign up he will get confuse of how to get there, because there would only be a signin button. We can wait with that issue for now though! I see that we have some conflicts and can not rebase, if you fix that by

PR #81, opened 2020-03-27
Feat/mission details

mat10tng:


src/app/page/Mission/Mission.js
```diff
@@ -0,0 +1,60 @@
+import React from "react";
```
azmainamin: minor: rename to Mission.jsx to be consistent

src/app/page/Missions/Missions.tsx
```diff
@@ -12,7 +12,7 @@ type Mission = {
   details: string;
```
azmainamin: Rename Missions to MissionsPage to remove confusion with Mission.js

PR #91, opened 2020-03-28
Link newly created missions to ownder

azmainamin:
When a logged user creates a new request/mission, we add them as owner to the mission document. We also add a mission entry in the user document to reflect how many and what missions they have created.

Notes:
- I had to use the uuid package to create mission uid. This means that we are not relying on Firebase to create a uid for mission any more. I had to do this because to save the mission Id on the user doc, I needed access to the mission Id, which, if we used Firebase's default uid, is hidden at that point.
- Refactor: I refactored some code on User model. We should check for regressions when a user clicks to volunteer for a mission.


mat10tng:
Ok, looks good. I think I will create more function so that we can use that, instead of using our own query, its getting really muddy with that in frontend

PR #95, opened 2020-03-29
Since data in missions from firestore save as firestore/ordered/missi…

mat10tng:
…ons. The problem lie in that this data does not clean up untill a firestore request have been sent -> stale data that display in the page, solution, save the data as different data set, such as missionsTodo, missionsVoluteered


thedma04:
There are conflicts with these PR

PR #105, opened 2020-04-01
user status

thirdeyeclub:
+new page added
+new nav link added

user can now set their status


src/App.jsx
```diff
@@ -82,9 +83,8 @@ function App() {
                 </PrivateRoute>
                 <Route path="/missions/:id" component={MissionDetails} />
                 <Route path="/missions" component={Missions} />
-                <PrivateRoute path="/user">
-                  <UserProfile />
-                </PrivateRoute>
+                <Route path="/user" component={UserProfile} />
```
aamin21, line 86: Might be a bigger discussion, but should the route be something like /user/:userId ? @mat10tng @utunga ?

src/app/page/Status/Status.jsx
```diff
@@ -0,0 +1,90 @@
...
+
+const StyledButton = styled(Button)`
+  margin-top: 24px;
+  flex-grow: 1;
+  width: 20vw;
+`;
+
+const Status = ({ firestore }) => {
+  // a happy little message
+  const [Thank, setThank] = useState(false);
```
aamin21, line 34: minor: rename Thank to thank just to be consistent with variable naming. Not a blocker obviously.

src/app/page/Status/Status.jsx
```diff
@@ -0,0 +1,90 @@
...
+      <StyledHeader variant="h1">
+        {" "}
+        Manually set your status so others know what you are doing.{" "}
+      </StyledHeader>
+      <StyledCopy variant="h2"> I am... </StyledCopy>
+      <StyledButton
+        color="primary"
+        size="large"
+        variant="contained"
+        onClick={() => setStatus("OnCall")}
```
aamin21, line 59: Minor: create a handleStatus() instead of creating a new function everytime the component is rendered. 
`handleStatusChange(){ setStatus("OnCall")}
<StyledButton onClick={handleStatusChange} />
`

src/App.jsx
mat10tng: I am going to add another route soon, so that route is going to be /user/profile later

aamin21: awesome! wanted to verify that.

src/app/page/Status/Status.jsx
thirdeyeclub: alright fixed

PR #106, opened 2020-04-02
See if user is signing in for the first time

azmainamin:
This is the first PR for US#60.
feature: figure out if the user is signing in for the first time.
refactor: move some firebase specific logic into a separate file than the LoginPage component.

### DIFF
--- a/src/app/page/Login/firebaseLogin.js
+++ b/src/app/page/Login/firebaseLogin.js
@@ -1,0 +1,45 @@
+export function loginWithSMS(firebase, phoneNumber) {
+  firebase.auth().useDeviceLanguage();
+  window.recaptchaVerifier = new firebase.auth.RecaptchaVerifier("recaptcha-container");
+
+  window.recaptchaVerifier.render().then(function (widgetId) {
+    window.recaptchaWidgetId = widgetId;
+  });
+  firebase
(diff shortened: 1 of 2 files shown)


src/app/page/Login/firebaseLogin.js
```diff
@@ -0,0 +1,45 @@
...
+  return firebase.login({ provider: "google", type: "popup" });
+}
+
+/**
+ * The idea is that if a user is signing in for the first time, the createdAt timestamp
+ * and lastSignIn timestamp will be very close (if not the same).
+ * The buffer is there to provide some flexibility i.e. network latency etc.
+ * @param {*} user
+ */
+export function firstTimeSignIn(user) {
```
mat10tng, line 42: Yeah, hahaha, nice idea! I am thinking of that this calculation would have to be done everytime an user login, so maybe we can set an isNewUser field instead when create a new user, and then set that to false?

azmainamin: @mat10tng I thought about that...but that means that we have to do this again, and set it to false on 2nd login. And then we will have to check if isFirstTime is true or false. I think that has more potential for bugs (forget to set it back to false or erroneously set it to false).

PR #118, opened 2020-04-05
look up function

thirdeyeclub:
created new function that looks up address and returns a object with lat and long

Take a look at it let me know if it needs to be imporved


src/app/page/MakeRequest/RequestForm.js
```diff
@@ -25,6 +25,27 @@ function RequestForm({ handleChange, values, onSubmit, getFile }) {
           label="Task Details"
         />
         <Header>Where do you need help?</Header>
```
utunga, line 27: At some point, we'll change this to 'delivery address' i guess

thirdeyeclub: I suppose yes

PR #120, opened 2020-04-05
Geo location to plot peoples places on map

thirdeyeclub:
! HEADS UP THIS PR WILL BREAK ANY MISSIONS WITHOUT ADDRESSES
PLEASE LET ME KNOW IF THERE ARE ANY PROBLEMS

I HAVE HAD SOME ISSUES WITH MAP BOX GIVING WEIRD LOCATION DATA TO ME SO WE MAY IN THE FUTURE WANT TO FIND BETTER GEOLOCATION SERVICE.


src/app/component/MapView/MapView.js
```diff
@@ -3,10 +3,11 @@ import { Map, TileLayer, Marker, Popup } from "react-leaflet";
 import { MapStyle } from "./MapView.style";
 // import Routing from './MapView-routing'
 
-function MapView() {
+function MapView(props) {
+  console.log(props)
```
mat10tng, line 7: remove this

thirdeyeclub: Strange I do not have this on my end, once i make another commit it will be gone.

package.json
```diff
@@ -16,6 +17,7 @@
     "@types/react-redux": "^7.1.7",
     "@types/react-router-dom": "^5.1.3",
     "awesome-typescript-loader": "^5.2.1",
+    "axios": "^0.19.2",
```
qdozaq, line 20: Is this a peer dependency for mapbox? I don't see it used anywhere else.
If it's not we should remove it.

We should be careful not to add packages for single use things as it can quickly bloat the app and make the site slower due to larger bundle sizes

src/app/hooks/addressLookUp.js
```diff
@@ -0,0 +1,27 @@
+const mbxClient = require("@mapbox/mapbox-sdk");
```
qdozaq, line 1: Lets make a `util` folder for this to go in instead as it's not really a hook and could be confusing

PR #122, opened 2020-04-06
Link Phone before volunteering. 

azmainamin:
feat: if someone who doesn't have a phone number wants to volunteer we force them to authenticate their number. The UX is as follows:

Sign In with Google/FB -> Try to Volunteer -> Cant because you do not have phone number -> Simple form to authenticate phone -> Happy Path: We link your phone auth and your non phone auth together. Sad Path: You already created an account with Phone number before, so we cannot link these two different accounts yet. We prompt you to sign in with that account.

### DIFF
--- a/src/app/model/User.js
+++ b/src/app/model/User.js
@@ -39,6 +39,24 @@ class User {
       });
   }
 
+  linkPhoneAuthentication(firebase, phoneNumber, recaptchaVerfier, callback) {
+    return firebase
+      .auth()
+      .currentUser.linkWithPhoneNumber(phoneNumber, recaptchaVerfier)
+      .then(function (confirmationResult) {
+        var code = window.prompt("Provide your SMS code");
+        recaptchaVerfier.clear();
+        return confirmationResult.confirm(code).then(() => {
+          callback();
+        });
+      })
+      .catch((err) =>
+        alert(
+          "You already have an account associated with this phone number. Please sign in using that number."
+        )
+      );
+  }
+
   getAuth = (state) => get(state, "firebase.auth");
 }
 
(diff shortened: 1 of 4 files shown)


src/app/model/User.js
```diff
@@ -39,6 +39,24 @@ class User {
...
+      .then(function (confirmationResult) {
+        var code = window.prompt("Provide your SMS code");
+        recaptchaVerfier.clear();
+        return confirmationResult.confirm(code).then(() => {
+          callback();
+        });
+      })
+      .catch((err) =>
+        alert(
+          "You already have an account associated with this phone number. Please sign in using that number."
```
mat10tng, line 55: can this situation even happen? 
User have an account -> he is logged in 
If he is logged in, and have phone number associate with the account -> he has a phone number

How can we come to this alert?

src/app/page/Missions/Missions.jsx
```diff
@@ -26,9 +28,31 @@ const PlaceHolder = styled.div`
...
+      User.linkPhoneAuthentication(firebase, phoneNumber, appVerifier, refreshPage);
+    }
+  }
+  function handlePopupClose() {
+    setPopupOpen(false);
+  }
   function volunteerForMission(missionId) {
-    User.assignAsVolunteer(firestore, missionId, user.uid);
+    if (!user.phoneNumber) {
+      setPopupOpen(true);
```
mat10tng, line 52: Actually, I have done some linking work already in the user profile page. Maybe instead of doing a popup just redirect them to user/profile if they are login but not have a number and say: Hey, add your phone number here

If they are not login, then -> login page

PR #124, opened 2020-04-06
Feat/user profile second iteration

mat10tng:


src/app/page/UserProfile/UserProfile.js
```diff
@@ -106,78 +138,75 @@ const UserProfile = ({ history, ...props }) => {
 
   async function errorHandler(error) {
```
aamin21, line 139: We should extract this out later, so that we can reuse this merge logic at other places: i.e. authenticate phone before volunteering.

mat10tng:
Create good first issues out of some of this later

PR #132, opened 2020-04-07
add loading spinner for pages with mission loading

malignantz:
Added MUI spinner to Missions, MissionsCreated and MissionsVolunteered pages.

### DIFF
--- a/src/app/page/Missions/Missions.jsx
+++ b/src/app/page/Missions/Missions.jsx
@@ -3,7 +3,7 @@ import { useFirestore, firestoreConnect } from "react-redux-firebase";
 import { useSelector, connect } from "react-redux";
 import { withRouter, Link } from "react-router-dom";
 
-import { Typography, Button, Grid } from "@material-ui/core";
+import { Typography, Button, Grid, CircularProgress } from "@material-ui/core";
 import styled from "styled-components";
 
 import { Page, Card } from "../../layout";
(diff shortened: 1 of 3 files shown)


src/app/page/Missions/Missions.jsx
```diff
@@ -39,9 +39,12 @@ const MissionsPage = ({ user, history, firebase, ...rest }) => {
       User.assignAsVolunteer(firestore, missionId, user.uid);
     }
   }
-
-  if (!missions) {
-    return <div> isloading...</div>;
+  if (missions === undefined) {
+    return (
+      <Grid container justify="center" alignItems="center" style={{ minHeight: "100vh" }}>
```
azmainamin, line 44: We can refactor the spinner + grids into a component so that we can reuse it at other places!

src/app/page/Missions/Missions.jsx
```diff
@@ -39,9 +39,12 @@ const MissionsPage = ({ user, history, firebase, ...rest }) => {
       User.assignAsVolunteer(firestore, missionId, user.uid);
     }
   }
-
-  if (!missions) {
-    return <div> isloading...</div>;
+  if (missions === undefined) {
```
azmainamin, line 42: Any particular reason for doing explicit check against undefined versus truthy check? Moreover, isn't missions an array? Would it be better to check for length then?

malignantz: Missions is undefined before data comes in, so I believe checking the length property would not work.

azmainamin: Gotcha!

src/app/page/Missions/Missions.jsx
azmainamin: Any thoughts on this? @malignantz

azmainamin:
There was another PR for the same issue that got merged. Closing this now.

Issue #137, opened 2020-04-07
Increase spacing between icon and text in the sidebar


azmainamin:
Fix spacing around icon and text:
![image](https://user-images.githubusercontent.com/12126548/78671542-707ae900-78ad-11ea-83ee-9e0168ae526f.png)

This is a nice mockup:
![image](https://user-images.githubusercontent.com/12126548/78671634-88526d00-78ad-11ea-884a-8c37bc8a64d5.png)

skbrown333:
It looks like I am unable to assign myself to this issue @utunga

nickx720:
Hi @skbrown333 , I have already fixed it in local but unable to push it 🤣

PR #144, opened 2020-04-07
Fixed the navigation menu alignment issue as per ticket 137

nickx720:
I have added the fix as requested by the issue tracker https://github.com/factn/resilience-app/issues/137

### DIFF
--- a/src/app/layout/NavigationDrawer/NavigationDrawer.js
+++ b/src/app/layout/NavigationDrawer/NavigationDrawer.js
@@ -2,7 +2,7 @@ import React from "react";
 import clsx from "clsx";
 import Drawer from "@material-ui/core/Drawer";
 import Button from "@material-ui/core/Button";
-import { useStyles } from "./NavigationDrawer.style";
+import { useStyles, iconStyle } from "./NavigationDrawer.style";
 import List from "@material-ui/core/List";
 import ListItem from "@material-ui/core/ListItem";
 import ListItemText from "@material-ui/core/ListItemText";
@@ -15,6 +15,7 @@ import { Link, useHistory } from "react-router-dom";
 import { useFirebase } from "react-redux-firebase";
 
 import { PrivateComponent } from "../../component";
+import { ListItemIcon } from "@material-ui/core";
 
 export default function TemporaryDrawer() {
   const firebase = useFirebase();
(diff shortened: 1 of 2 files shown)


src/app/layout/NavigationDrawer/NavigationDrawer.style.js
```diff
@@ -25,3 +25,7 @@ export const useStyles = makeStyles({
     color: "#150E60",
   },
 });
+
+export const iconStyle = {
```
mat10tng, line 29: this should go into the useStyles

src/app/layout/NavigationDrawer/NavigationDrawer.js
```diff
@@ -15,6 +15,7 @@ import { Link, useHistory } from "react-router-dom";
 import { useFirebase } from "react-redux-firebase";
 
 import { PrivateComponent } from "../../component";
+import { ListItemIcon } from "@material-ui/core";
 
 export default function TemporaryDrawer() {
   const firebase = useFirebase();
```
mat10tng, line 21: I think this firebase is not used right now, maybe we can remove some unused import

src/app/layout/NavigationDrawer/NavigationDrawer.style.js
nickx720: Sure @mat10tng , I was looking at material-ui documentation and they haven't specified a use case with useStyles. https://material-ui.com/components/icons/ 
I thought instead of writing it inline I could offload it to the style.js. What are your thoughts?

src/app/layout/NavigationDrawer/NavigationDrawer.js
nickx720: Sure. I can do that! ☺️

src/app/layout/NavigationDrawer/NavigationDrawer.style.js
mat10tng: This is because of consistency tbh, just so that it would make it easier for other to read

src/app/layout/NavigationDrawer/NavigationDrawer.js
nickx720: @mat10tng I have updated styles to useStyles and there is a logout feature which is being used by firebase for signout so I am unable to remove it.

Issue #138, opened 2020-04-07
Make the About text bigger and add spacing around it


azmainamin:
Tldr: Prettify it. Maybe a small button?

![image](https://user-images.githubusercontent.com/12126548/78671923-ee3ef480-78ad-11ea-8b8f-b1d84ff996f4.png)

ShanuCreations:
@azmainamin - Could you please assign this issue to me as I've fixed this locally and the change is ready to push ?

PR #145, opened 2020-04-07
Fixed About text styling in the Home Page as per #138

ShanuCreations:
I've added a fix as per #138 issue, could someone please review ?


mat10tng:
Did you forgot the home.style ?

ShanuCreations:
Nope, I didn't had to change anything in home.style. I just used the existing 'StyledHomeButton' to achieve the attached result. Hope this is the expected result, let me know otherwise.

New About Button

Issue #111, opened 2020-04-02
[BUG]User goes to a "cannot get" route page instead of login page when entering a protected route in the search bar


CSumm:
Steps to reproduce

1. Open[ the heroku app](https://mutualaidworld-frontend.herokuapp.com/)
2. Enter a protected route inside of the search bar
3. Observe the output of the page.

Expected result:
User can be redirected to login page with the warning "You must be logged in".

Actual result:
User is greeted to a "cannot get" route page instead of the login page that specifies a login warning.

AquiburRahmanKhan:
@utunga can you please assign this issue to me ? thanks

PR #146, opened 2020-04-07
fix: reroute unauthorized user to login page when user tries to acces…

AquiburRKhan:
…s private route


AquiburRKhan:
I've added a fix as per #111 issue, could someone please review ?

Issue #121, opened 2020-04-05
[UI BUG] We need a spinner or something to show data is loading (for example on mission page)


utunga:
Currently, just for example, the screen which lists all current missions is 'just white' while missions load. Then, if there are no missions it stays white.

We need some sort of loading spinner while it's loading. We could also do with something like a '0 missions matching this criteria' or something if it resolves with 0 missions.

PR #147, opened 2020-04-07
Loading Spinners

jwu910:
Add loading spinners to Missions pages at the list level so the page header is still visible. Added in a style injection for MUI circular progress component at the app level to center circular progress components. Please let me know if you have any questions. #121

### DIFF
--- a/src/app/page/Missions/Missions.jsx
+++ b/src/app/page/Missions/Missions.jsx
@@ -3,7 +3,7 @@ import { useFirestore, firestoreConnect } from "react-redux-firebase";
 import { useSelector, connect } from "react-redux";
 import { withRouter, Link } from "react-router-dom";
 
-import { Typography, Button, Grid } from "@material-ui/core";
+import { CircularProgress, Typography, Button, Grid } from "@material-ui/core";
 import styled from "styled-components";
 
 import { Page, Card } from "../../layout";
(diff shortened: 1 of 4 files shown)


src/app/page/Missions/Missions.jsx
```diff
@@ -40,41 +40,41 @@ const MissionsPage = ({ user, history, firebase, ...rest }) => {
...
+                  history.push(`/missions/${mission.id}`);
+                }}
+              >
+                Details
+              </StyledButton>
+            </Grid>
+          </Card>
+        ))
+      ) : (
+        <CircularProgress />
```
thedma04, line 76: This is good, but in a situation where there are errors fetching missions, the loading will be there forever until refreshed. Do we wanna keep a loading and error state to mitigate this behaviour??

src/app/page/MissionsCreated/Missions.jsx
```diff
@@ -25,24 +25,28 @@ const MissionsPage = ({ auth, history, firebase, ...rest }) => {
...
+                  history.push(`/missions/${mission.id}`);
+                }}
+              >
+                Details
+              </StyledButton>
+            </Grid>
+          </Card>
+        ))
+      ) : (
+        <CircularProgress />
```
thedma04, line 48: The same applies to this

thedma04, line 48: The same applies to this

src/app/page/Missions/Missions.jsx
azmainamin: I think we have to.  Maybe create a separate issue for that?

jwu910: Correct, Felt that any error handling would creep the scope of this ticket. Was planning to bring that up in another issue.

The missions object looks to be assigned through a useSelector hook(?) that I'm I hadn't had a chance to get familiar with yet.

Would you advise to change the scope of the ticket to include error handling for the api call?

thedma04: Can you create an issue for that whilst I approve this

PR #151, opened 2020-04-07
dont reload, route location exist only in vrtl dom

pedro-surf:


src/app/page/UserProfile/LinkPhoneAccount.jsx
```diff
@@ -32,7 +32,7 @@ function LinkPhoneAccount({ firebase, auth, data, errorHandler }) {
       await confirmationResult.confirm(verificationCode);
       verifier.clear();
       setSuccessSnackbarOpen(true);
-      window.location.reload(false);
+      //window.location.reload(false);
```
azmainamin, line 35: So the idea was to refresh the page, so that the User Profile page reflects the change. Can we do a force refresh instead of commenting it out?

azmainamin:
Per our conversation, the actual problem lies in merging account. Closing now.

PR #153, opened 2020-04-08
Fix bug when changing phone number

pedro-surf:


src/app/page/UserProfile/LinkPhoneAccount.jsx
```diff
@@ -16,23 +16,51 @@ function LinkPhoneAccount({ firebase, auth, data, errorHandler }) {
   const currentUserPhoneNumber = data?.phoneNumber || "";
   const [phoneNumber, updatePhoneNumber] = useState(currentUserPhoneNumber);
   const [successSnackbarOpen, setSuccessSnackbarOpen] = useState(false);
+  const [userHasPhone, setUserHasPhone] = useState(false);
   const classes = useStyles();
 
+  useEffect(() => {
```
qdozaq, line 22: since `""` evaluates to false we can simply use `currentUserPhoneNumber` as the flag to check and not have to use this extra hook/state.

pedro-surf: srry I was doing rebase. @qdozaq had to force a push

pedro-surf: But how to check it when firing the change button? Because by then, it won't be "" anymore

qdozaq: `currentUserPhoneNumber` is just coming from the `data` prop. That would only change once we submit the new number. `phoneNumber` is the one thats getting updated on user input

pedro-surf: @qdozaq u're right. I'll fix it, thanks

pedro-surf: @qdozaq Done. However @azmainamin said they've tried hard and couldn't make this prop update itself and the component. For that reason there's a window reload call, feel free to dive into it. (I'll go sleep, 11PM here :sleeping:)

qdozaq: Thats odd. but as long as it's not breaking anything I guess it's fine for now.
Wonder if something with firebase not triggering a react state change

src/app/page/UserProfile/LinkPhoneAccount.jsx
```diff
@@ -19,20 +19,38 @@ function LinkPhoneAccount({ firebase, auth, data, errorHandler }) {
...
+      } else {
+        const provider = new firebase.auth.PhoneAuthProvider();
+        const verificationId = await provider.verifyPhoneNumber(phoneNumber, verifier);
+        const verificationCode = window.prompt(verificationPrompt);
+        const phoneCredential = firebase.auth.PhoneAuthProvider.credential(
+          verificationId,
+          verificationCode
+        );
+        await auth.currentUser.updatePhoneNumber(phoneCredential);
+        successCallback(verifier);
```
qdozaq, line 52: could move this callback out of the if/else so it only has to be called in the one place.
Think this looks good otherwise.

azmainamin:
@pedro-surf Everything looks good. I checked the data in Firebase as well. Approved and merged! The issue should be close as well.

Issue #142, opened 2020-04-07
Add jsDoc for functions


azmainamin:
Code to cover:

• src/app/model
• src/app/hooks/useForm

laredotornado:
Hi, When I get added to the project, I would like to volunteer to take this issue.  Thanks, - Dave

PR #154, opened 2020-04-08
Added jsdocs to userForm.js and app/models 

laredotornado:
This is the PR for issue https://github.com/factn/resilience-app/issues/142


package.json
```diff
@@ -4,19 +4,22 @@
   "private": true,
   "dependencies": {
     "@firebase/auth-types": "^0.10.0",
-    "@material-ui/core": "^4.9.7",
+    "@material-ui/core": "^4.9.9",
```
qdozaq, line 7: Yeh none of these packages should need to updated since we're using `^`.
Looks like something may have auto updated these?

Can you revert?

package.json
```diff
@@ -4,19 +4,22 @@
...
-    "@types/react": "^16.9.25",
+    "@types/react": "^16.9.32",
     "@types/react-redux": "^7.1.7",
     "@types/react-router-dom": "^5.1.3",
     "awesome-typescript-loader": "^5.2.1",
+    "better-docs": "^1.4.7",
     "express": "^4.17.1",
-    "firebase": "^7.12.0",
+    "firebase": "^7.13.2",
+    "jsdoc": "^3.6.4",
```
qdozaq, line 22: You should add this to `devDependencies`.
Probably best to leave the other packages the way they are unless there's a specific patch you need?

laredotornado: Oops, my bad.  Reverted.

Issue #156, opened 2020-04-08
[BUG] Error w/ Mission Detail Volunteer Button


Spraynard:
The following bug is seen on the master branch while running a local copy. 
```
TypeError: _model__WEBPACK_IMPORTED_MODULE_10__.User.assginedToMission is not a function
volunteerForMission
src/app/page/MissionDetails/MissionDetails.jsx:48
```

This is also seen on the Heroku demo but it is not as flagrant of a fail (shows error in console)

### Steps to Reproduce

1. Start up the local copy
1. Perform the following actions while *not* logged in
1. Go to `/missions`
1. Click on **Details** button of any job to go to the `/missions/{mission_id}` page
1. Click on **Volunteer** button

### Expected Functionality

I would expect it to bring up the **Add a phone number** dialog that comes up when you don't have a phone number added as is seen on the `/missions` page when you click on **Volunteer**

### Notes
I am new to this repo so I'm a little unaware of the app workflow. So, I'm unsure if a user account is needed to be created for this

azmainamin:
I believe this can be close now, right @mat10tng ?

mat10tng:
yes.

PR #157, opened 2020-04-08
Update MissionDetail to show UserPhoneUnverifiedPopup

Spraynard:
Applies to #156. 

Saw that MissionDetail.jsx was using `User.assignedToMission` in its `volunteerForMission()` function. `assignedToMission` was not defined.

So, I structured it to look similar to user mission assigning procedures in `Missions.jsx`

### DIFF
--- a/src/app/component/UserPhoneUnverifiedPopup.jsx
+++ b/src/app/component/UserPhoneUnverifiedPopup.jsx
@@ -1,0 +1,37 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+import Popup from "./Popup";
+
+import { Grid, Typography } from "@material-ui/core";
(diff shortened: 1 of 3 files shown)


src/app/component/UserPhoneUnverifiedPopup.jsx
```diff
@@ -0,0 +1,37 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+import Popup from "./Popup";
+
+import { Grid, Typography } from "@material-ui/core";
+import { Link } from "react-router-dom";
+
+/**
```
mat10tng, line 9: nice!

Spraynard: Thanks :) Glad u like

PR #165, opened 2020-04-08
Remove unused imports, fix dependency array in useEffect Closes #149

### LINKED ISSUE #149: Remove unused imports
When you start the application there are so much unused imports, I think we should get rid of them, because we can miss more important warnings, like memory leaks etc.

roberts98:
Removed all unused imports and vars + added missing dependency in useffect deps array.


roberts98:
Closes #149

PR #179, opened 2020-04-09
Feat/story-133 home page

mat10tng:
origin/master
home
this change will add

1. splash screen just for the homepage
2. modified change to the withLoading Hoc to accept splash screen
3.New Home page

splash
new home


src/app/page/Home/Home.js
```diff
@@ -4,62 +4,141 @@ import PropTypes from "prop-types";
...
-          <StyledHomeButton
-            onClick={() => history.push("/request/create")}
-            text="Request Help"
-            secondary
-          />
-          <StyledHomeButton onClick={() => history.push("/about")}>About</StyledHomeButton>
-        </>
+        <Grid container>
+          <Grid container>
+            <img src={HomeImage1} className={classes.HomeImage} alt="" />
```
azmainamin, line 85: We should add some alt tag to make aXe stop yelling.

src/app/page/Home/Home.js
```diff
@@ -4,62 +4,141 @@ import PropTypes from "prop-types";
...
+          </Grid>
+          <Grid container>
+            <Button className={classes.LoginButton} onClick={() => history.push("/login")}>
+              Log in
```
azmainamin, line 89: Can we rename to Signin instead of login?

src/app/page/Home/Home.js
mat10tng: this one is decorative and so have a alt=""  (empty string)

src/app/page/Home/Home.js
mat10tng: yeah

PR #184, opened 2020-04-09
when user adds address to their profile we save it

thirdeyeclub:
-Better data so we can see when what volunteers are in what area
- we need to update profile to display users info because it does not do that, you can only update it


package.json
```diff
@@ -9,12 +9,14 @@
...
     "@material-ui/lab": "^4.0.0-alpha.48",
     "@reduxjs/toolkit": "^1.3.4",
+    "ci": "^1.0.0",
     "express": "^4.17.1",
     "firebase": "^7.13.2",
     "jsdoc": "^3.6.4",
     "leaflet": "^1.6.0",
     "leaflet-routing-machine": "^3.2.12",
     "lodash": "^4.17.15",
+    "npm": "^6.14.4",
```
qdozaq, line 19: I don't think we need either of these packages

package.json
```diff
@@ -9,12 +9,14 @@
     "@material-ui/icons": "^4.9.1",
     "@material-ui/lab": "^4.0.0-alpha.48",
     "@reduxjs/toolkit": "^1.3.4",
+    "ci": "^1.0.0",
     "express": "^4.17.1",
```
azmainamin, line 13: Also, why do we have express? :/

src/app/page/UserProfile/UserOverview.jsx
```diff
@@ -15,12 +19,26 @@ const useStyles = makeStyles((theme) => ({
...
 
   const displayName = _.get(profile, "displayName", "");
   const address = _.get(profile, "address", "");
 
+  useEffect(() => {
+    const userId = user.uid;
+    if (address) {
+      const location = addressLookUp(address);
+      location
+        .then((res) => firestore.collection("users").doc(userId).set({ location: res }))
```
azmainamin, line 35: Does set() also update? I am guessing it does but wouldn't mind a verification.

src/app/page/UserProfile/UserOverview.jsx
```diff
@@ -1,7 +1,11 @@
-import React from "react";
+import React, { useEffect } from "react";
 import { H5 } from "../../component";
 import { Grid, Avatar, TextField } from "@material-ui/core";
 import { makeStyles } from "@material-ui/core/styles";
+import addressLookUp from "../../utils/addressLookUp";
+import { useSelector } from "react-redux";
+import { withRouter } from "react-router-dom";
+import { withFirestore } from "react-redux-firebase";
```
azmainamin, line 8: I prefer useFirestore() instead of withFirestore. It means 1 less prop PLUS will make testing a LOT easier.

thirdeyeclub:
> 
> 
> dude this is such a big change in package-lock

I see that, I have been having issues with npm big time. I will go over it and clean it out

package.json
thirdeyeclub: it appears Tuan added that when we first started the repo! that has been there for a while. I dont think we need it, but ill test that before removing it

thirdeyeclub:
Closing this until I can figure out issues with firebase

package.json
mat10tng: ehm, we need that for heroku

mat10tng: please dont remove express

src/app/page/UserProfile/UserOverview.jsx
mat10tng: set does not update, it actually set the entire profile, so this would be, yeah, not good

src/app/page/UserProfile/UserOverview.jsx
mat10tng: I have another suggestion, using firestoreConnect and get your props like this;

const mapStateToProps = (state, ownProps) => {
  const missionId = ownProps.match.params.id;
  const mission = state.firestore.data.missions && state.firestore.data.missions[missionId];
  return {
    auth: state.firebase.auth,
    missionId,
    mission,
  };
};
export default compose(
  connect(mapStateToProps),
  firestoreConnect((props) => {
    const missionId = props.missionId;
    if (!props.auth.uid) return [];
    return [
      {
        collection: "missions",
        doc: missionId,
      },
    ];
  })
)(MissionDetailsPage);

it allows more consistency for how you getting your stuff, I find useAnything... a bit too magical and hard to find

Issue #143, opened 2020-04-07
Add propTypes and jsDoc for components in Pages


azmainamin:
Use PropTypes(https://reactjs.org/docs/typechecking-with-proptypes.html) for components and JSDOCs(https://jsdoc.app/about-getting-started.html)for functions

These are nice to have. Code to cover:

• src/app/page
		• About Us

		• Home

		• Login

		• MakeRequest

		• MissonDetails

		• Missions

		• MissionsControl

		• MissionsCreated

		• MissionsVolunteered

		• OrganizerSignup

		• Signup

		• Status

		• UserProfile

guillermoparral1995:
has anyone taken this? if not, I'd like to have this one (and the other ones about propTypes if they are still available)

guillermoparral1995:
why was this closed?

mat10tng:
Oh sorry. I was on a cleaning spray

guillermoparral1995:
removed my assignment by accident 😅

azmainamin:
@guillermoparral1995  added you back!

guillermoparral1995:
This should be closed, PR was merged 😄

PR #186, opened 2020-04-09
Add propTypes and jsDoc for components in Pages

guillermoparral1995:
Issue: #143


package.json
```diff
@@ -76,6 +76,7 @@
     "better-docs": "^1.4.7",
     "husky": "^4.2.3",
     "jest-dom": "^3.5.0",
+    "jsdoc": "^3.6.4",
```
qdozaq, line 79: Thats strange. This is already on master. Can you rebase?

src/app/component/MapView/MapView.js
```diff
@@ -4,7 +4,7 @@ import { MapStyle } from "./MapView.style";
 // import Routing from './MapView-routing'
 
 function MapView(props) {
-  console.log(props)
+  console.log(props);
```
qdozaq, line 7: This should just be removed

package.json
qdozaq: And remove the changes from the package-json.lock

azmainamin:
There are also multiple conflicts that need to be resolved

package.json
guillermoparral1995: i think the problem is that I used it as a dev dependency but it was already installed as non-dev dependency. i think it should remain dev

qdozaq: Oh I thought this had be moved to dev. Yeh you can leave this one and remove the one from `dependencies` that'd be great.

package.json
```diff
@@ -15,7 +15,6 @@
     "date-fns": "^2.12.0",
     "express": "^4.17.1",
     "firebase": "^7.14.0",
-    "jsdoc": "^3.6.4",
```
qdozaq, line 18: Looks like this didn't get moved to `dev-dependencies`

guillermoparral1995: whoops, now it should be ok! 😅

PR #187, opened 2020-04-10
Add Create missions route

pedro-surf:
- Adding Create Missions route. Requests seem ok. 

- I've created my own account in Algolia Places and inserted into a 3rd party component to handle the location input. I hope that we can create a proper one for the project, however this one can be reused for now @utunga 

(I've replicated the existing CRUD function from MakeRequest. If it stops working there, we will have to fix both.)

screenshot: 
Screenshot from 2020-04-09 22-11-31


pedro-surf:
I've renamed the route, it was having a string conflict that made it unaccessible. Also inserted a link to it in the menu.

PR #192, opened 2020-04-10
Feat/merge accounts

mat10tng:


src/app/component/index.js
```diff
@@ -3,6 +3,7 @@ import Chip from "./Chip";
 import MissionCard from "./MissionCard";
 import PrivateComponent from "./PrivateComponent";
 import ProtectedRoute from "./PrivateComponent";
+import Popup from "./Popup";
```
azmainamin, line 6: this is for my own understanding: what is the purpose of adding Popup in this index file? So that we can do:
`import Popup from "../component"?

mat10tng: ah yes, just a reflex. I usually try to keep a two level import, one in the index, and one directly. I think the same strategy is done by material-ui

PR #195, opened 2020-04-10
194: Update MissionCard

### LINKED ISSUE #194: Standardize on one MissionCard.jsx component
We currently have two (very similar) versions of `MissionCard.js`.  One version is used by both the  _Volunteers Needed_ and _Volunteered Missions_ pages.   The other type is used by MissionControl.

We should combined those into one component because DRY.  

[See comment.](https://github.com/factn/resilience-app/pull/193#pullrequestreview-391672495)

Eihcir0:
@qdozaq After doing this, I'm not sure if it was a better idea to do it this way or maybe leave the two components since they are used by different things and just rename the other `MissionCard.jsx` => `MissionControlCard.jsx` or something.  

closes #194


mat10tng:
oh no, ehm, you can just ahead and remove the mission card section that belong to the missions control, it was a private component

Eihcir0:
@mat10tng Ahh that clears it up. @qdozaq Let's just close this and consider renaming that private component in the future...

PR #203, opened 2020-04-11
Add food-box mission route

pedro-surf:
Add new route for organizers create food box mission.

### DIFF
--- a/src/app/page/MakeFoodBoxMission/MakeFoodBoxMission.jsx
+++ b/src/app/page/MakeFoodBoxMission/MakeFoodBoxMission.jsx
@@ -1,0 +1,81 @@
+import React, { useState } from "react";
+import CircularProgress from "@material-ui/core/CircularProgress";
+import { User } from "../../model";
+import { withRouter } from "react-router-dom";
+import "firebase/storage";
+import MissionForm from "./MissionForm";
+import useForm from "../../hooks/useForm";
+import { useSelector } from "react-redux";
+import { getFirebase, withFirestore } from "react-redux-firebase";
+import { v4 as uuidv4 } from "uuid";
+
+function MakeFoodBoxMission({ history, firestore }) {
+  const firebase = getFirebase();
+  const [loading, setLoading] = useState(false);
+  const [file, setFile] = useState(null);
+  // const [autoAssignHelper, setAutoAssignHelper] = useState(false);
+  const storage = firebase.storage();
+  const { handleChange, values, setValues } = useForm();
+
+  const user = useSelector((state) => state.firebase.auth);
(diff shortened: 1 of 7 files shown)


src/app/page/MakeFoodBoxMission/MakeFoodBoxMission.jsx
```diff
@@ -0,0 +1,81 @@
...
+   */
+  function saveMissions(payload) {
+    console.log("Saving mission...");
+    const model = { ...payload, ownerId: user.uid, type: "foodBox" };
+    const missionId = uuidv4();
+
+    firestore
+      .collection("missions")
+      .doc(missionId)
+      .set({ ...model, status: "todo" });
```
mat10tng, line 37: we could, just update what need to be update instead of set here

src/app/page/MakeFoodBoxMission/MissionForm.js
```diff
@@ -0,0 +1,246 @@
...
+        <MuiPickersUtilsProvider utils={MomentUtils}>
+          <StyledHeader align="left" variant="h2">
+            Pickup
+          </StyledHeader>
+          <AlgoliaPlaces
+            placeholder="Pickup location"
+            name="pickUp"
+            options={{
+              appId: "plZ318O8ODTC",
+              apiKey: "b5e0781d289a9aa8edb37bf24aef874e",
```
qdozaq, line 153: We should break these out into consts and share between the components

src/app/page/MakeFoodBoxMission/Request.style.js
```diff
@@ -0,0 +1,22 @@
+import styled from "styled-components";
+import ImageUpload from "../../component/ImageUpload";
+
+import { makeStyles } from "@material-ui/core/styles";
+
+export const useStyles = makeStyles((theme) => ({
```
qdozaq, line 6: For future reference we're passing the same theme to the styled components ThemeProvider so you should be able to access the theme there too if you prefer

```js
styled.div`
margin: ${(props) => props.theme.spacing(1)}
`
```

src/app/page/MakeFoodBoxMission/MissionForm.js
pedro-surf: Thanks a lot for reviewing. I'll keep a single [PR](https://github.com/factn/resilience-app/pull/211) , and there I shall correct what you guys have pointed. I think FoodBox mission can be just a type to be chosen, since it's part of the same collection, we don't need a separate route. @mat10tng @utunga thoughts?

pedro-surf:
According to the current flow, there is no need for a separate FoodBox mission route

PR #205, opened 2020-04-12
Initialize firebase cloud functions

qdozaq:
This is just the base initialization from running `firebase init functions`

Lookin to use these for paypal integrations.

### DIFF
--- a/functions/.gitignore
+++ b/functions/.gitignore
@@ -1,0 +1,8 @@
+## Compiled JavaScript files
+**/*.js
+**/*.js.map
+
+# Typescript v1 declaration files
+typings/
+
+node_modules/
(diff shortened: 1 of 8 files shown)


functions/package-lock.json
```diff
@@ -0,0 +1,2388 @@
+{
```
azmainamin, line 1: Should the lock file be this big?

functions/.gitignore
```diff
@@ -0,0 +1,8 @@
+## Compiled JavaScript files
+**/*.js
```
azmainamin, line 2: Just to verify, should we ignore .js files as well?

qdozaq: This glob should handle everything including *.js

functions/package-lock.json
qdozaq: Yeh this is just from the default firebase and typescript modules. I guess they just have hella dependencies

PR #214, opened 2020-04-12
Updating Mission List to be more like Figma Spec

Spraynard:
### Summary

Updated `MissionCard` to accept children (i.e. our buttons) and also perform more like a `Card` component

Updated overall `MissionCard` styling per the spec on Figma

### Notes
Before
Screen Shot 2020-04-12 at 3 51 51 PM

After
Screen Shot 2020-04-12 at 3 50 31 PM


Spraynard:
Here is the figma spec that I referenced https://www.figma.com/file/v5HdxDsvO2NzMuZU6CArzF/Resilience-App?node-id=2784%3A4607

Issue #136, opened 2020-04-07
aXe console error 


azmainamin:
Going to SignIn page gives this:
![image](https://user-images.githubusercontent.com/12126548/78671191-f480a100-78ac-11ea-812c-7cf9f99fc985.png)

azmainamin:
This is propagated across the whole app.

mat10tng:
Yeah, I have no clue about this bug. From what I have seen, I am thinking of it is how React add and remove node from the dom -> but Axe did not catch that and wonder, where did the damn node go, and so we got a weird error coming out.

sebassdc:
@azmainamin i think this have to do with Transition Group, i think react-axe is not able to keep up with this.
I was  researching a way to filter this error so this don't show up in the console but didn't find anything.

sebassdc:
Also there are some console errors due to the svg using identical ids because we are using the same svg twice in the home page.

heisfullstacked:
Can I have a look into this?

mTRx0:
@mat10tng @azmainamin I guess this issue can be closed as it seems to be fixed.

PR #215, opened 2020-04-12
Enhancement/fix console errors warnings

gabz75:
Hello 👋 

This PR includes a fix for https://github.com/factn/resilience-app/issues/136 and also fixes all the `console.error`s that I was able to find. I also fixed the tests.

Also, I added a github action to run the tests and linting. We could run `prettier --check` to enforce consistency?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,27 @@
+name: resilience-app CI
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [10.x]
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+      - run: npm ci
+      - run: npm test
+      - run: npm run lint
+        env:
+          CI: true
--- a/package.json
+++ b/package.json
@@ -43,7 +43,8 @@
     "eject": "react-scripts eject",
     "storybook": "start-storybook -p 9009 -s public",
     "build-storybook": "build-storybook -s public",
-    "docs": "jsdoc -c jsdoc.conf.json"
+    "docs": "jsdoc -c jsdoc.conf.json",
+    "lint": "./node_modules/.bin/eslint . --ext .js,.jsx,.ts,.tsx"
   },
   "eslintConfig": {
     "extends": "react-app"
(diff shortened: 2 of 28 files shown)


package.json
```diff
@@ -83,7 +84,7 @@
     "jsdoc": "^3.6.4",
     "lint-staged": "^10.1.2",
     "prettier": "^2.0.4",
-    "react-axe": "^3.4.1",
+    "react-axe": "3.3.0",
```
gabz75, line 87: See: https://github.com/dequelabs/react-axe/issues/74#issuecomment-599754902

It looks like many folks are experiencing similar issues with `react-axe@3.4.1`. In `3.3.0`, the error disappears.

src/app/component/ImageUpload/ImageUpload.test.js
```diff
@@ -1,11 +1,8 @@
 import React from "react";
-import { render, cleanup, fireEvent } from "@testing-library/react";
-import "jest-dom/extend-expect";
+import { render, fireEvent } from "@testing-library/react";
 import { ImageUpload } from "./ImageUpload";
 
 describe("ImageUpload Component", () => {
-  afterEach(cleanup);
```
gabz75, line 7: This is automatically setup with Jest, see: https://testing-library.com/docs/react-testing-library/api#cleanup

src/app/component/ImageUpload/ImageUpload.test.js
```diff
@@ -1,11 +1,8 @@
 import React from "react";
-import { render, cleanup, fireEvent } from "@testing-library/react";
-import "jest-dom/extend-expect";
```
gabz75, line 3: Already defined in `setupTests.js`

src/app/component/MapView/MapView.js
```diff
@@ -12,7 +12,7 @@ function MapView(props) {
   });
   const position = [state.lat, state.lng];
   return (
-    <Map center={position} zoom={state.zoom} style={MapStyle} data-testid="map">
+    <Map center={position} zoom={state.zoom} style={MapStyle} id="leaflet-map">
```
qdozaq, line 15: This could be problematic if we want to have multiple MapViews on one page.

azmainamin: Maybe use a bit more unique id: leaflet-missions-map or something like that. I suck at naming.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,30 @@
+# This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
```
azmainamin, line 1: This is awesome. thanks for doing this!

src/app/component/MapView/MapView.js
gabz75: **Update: React.Fragment only supports {key} and {children}, so defaulting back to `div`**

Good point, `react-leaflet` doesn't support `data-testid` unfortunately. Since this is a selector for tests, another solution would be to wrap everything with a `<div>` so we can use `getByTestId` in the tests.
```
<div data-testid="leaflet-mapview">
  <Map ... />
</div>
```

qdozaq: Hmm would leaflet work with a class name?

Issue #216, opened 2020-04-13
Fix dashboard drawer UI bug - @dvalinotti


azmainamin:
The bug:

![image](https://user-images.githubusercontent.com/12126548/79083503-aa5c4d00-7cfc-11ea-891d-14abda71a95a.png)

Opening the drawer should push the items to the right instead of covering them.

azmainamin:
IN PROGRESS BY: @dvalinotti

PR #218, opened 2020-04-13
Resolve page content shifting bug on /dashboard

dvalinotti:
https://github.com/factn/resilience-app/issues/216

On /dashboard, main page content would not shift to the right when the Drawer menu is open.
Resolved by passing className as a prop to the Home component that updates when the drawer "open" state variable is changed to shift the content to the right.

### DIFF
--- a/src/app/dashboard/Appbar.jsx
+++ b/src/app/dashboard/Appbar.jsx
@@ -31,10 +31,10 @@ const useStyles = makeStyles((theme) => ({
     }),
   },
   menuButton: {
-    marginLeft: "4px",
+    marginLeft: theme.spacing(0.5),
   },
   toolBarTitle: {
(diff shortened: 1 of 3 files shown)


src/app/dashboard/Appbar.jsx
```diff
@@ -33,6 +33,9 @@ const useStyles = makeStyles((theme) => ({
   menuButton: {
     marginLeft: "4px",
   },
+  toolBarTitle: {
+    marginLeft: "1rem",
```
mat10tng, line 37: can we use theme.spacing for as much as possible ?

dvalinotti: I've updated that entry and other changes I've made to use theme.spacing instead of rem or px. Let me know if those changes look good!

PR #230, opened 2020-04-14
Paypal id settings

jonparker:
@qdozaq I hope I'm on the right track with this.
This is just a draft to make sure I don't go down the wrong path.

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -25,6 +25,7 @@ import {
   MissionsVolunteered,
   MakeMission,
   PostMission,
+  Settings,
 } from "./app/page";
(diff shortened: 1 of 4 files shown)


qdozaq:
Looks good so far. Obviously up for design to figure out, but we'll probably want something similar to the userprofile page where you can edit and save info which is sent to firebase

jonparker:
Ok, makes sense. I'll have a look at user profile as a reference.

jwu910:
Is this still being worked on?

qdozaq:
Nah we can close this

PR #237, opened 2020-04-15
Change text in MissionForm date pickers to display default date value

### LINKED ISSUE #233: Creating a new mission needs to default to today's date
When creating a new mission, although the "Pickup Time" is pre-populated with the current time, the "Select Date" field does not default to the current date ...

Screen Shot 2020-04-14 at 5 25 06 PM

Should default to today's date.

dvalinotti:
resolves #233

Cause of issue was the default value of pickUpDateLabel and dropOffDateLabel being null rather than the current date. State variables that stored the date pickers' values were correctly set to the current date/time of the user. Resolved the issue by setting the pickUpDateLabel and dropOffDateLabel default values to the current date/time stringified.


src/app/page/MakeMission/MissionForm.js
```diff
@@ -88,15 +92,15 @@ function MissionForm({ handleChange, values, onSubmit, getFile /*, assignHelper,
       return;
     }
     if (stage === "pickUp") {
-      setPickUpDateLabel(date);
+      setPickUpDateLabel(date.toString().substr(0, 15));
```
qdozaq, line 95: According to lines `97` and `107` looks like `data` could be undefined so may want to have a fallback on these sets as well

azmainamin:
Merge conflicts

src/app/page/MakeMission/MissionForm.js
dvalinotti: Good catch - Resolved this by adding a condition that checks if `date` is undefined before assigning values

dvalinotti:
> Merge conflicts

Resolved merge conflicts!

PR #239, opened 2020-04-15
Creating map points for users. related card 112

thirdeyeclub:
As a user when an address is added on their profile page it geolocates their address, adds lat/long to their account, and puts them on a map

### DIFF
--- a/src/app/page/UserProfile/UserProfile.js
+++ b/src/app/page/UserProfile/UserProfile.js
@@ -11,6 +11,7 @@ import { makeStyles } from "@material-ui/core/styles";
 
 import { useFirebase, useFirestore } from "react-redux-firebase";
 import _ from "lodash";
+import addressLookUp from "../../utils/addressLookUp";
 
 import LinkGoogleAccount from "./LinkGoogleAccount";
 import LinkPhoneAccount from "./LinkPhoneAccount";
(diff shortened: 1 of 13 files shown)


src/app/page/UserProfile/UserProfile.js
```diff
@@ -75,9 +76,39 @@ const UserProfile = ({ history }) => {
...
+      !profile.address
+        ? console.log(null)
+        : await addressLookUp(profile.address).then((data) =>
+            firebase.updateProfile({
+              ...profile,
+              addressMapPoint: { latitude: data.lat, longitude: data.long },
+            })
+          );
+    } catch (error) {
+      if (error.response) {
```
qdozaq, line 91: Does there need to be any fallback action if we reach this catch? Like telling the user something?

Lets also clean up these console.logs

azmainamin: I would use the ErrorSnackbar here.

src/app/page/UserProfile/UserProfile.js
```diff
@@ -75,9 +76,39 @@ const UserProfile = ({ history }) => {
...
     setView("view");
   }
-  function saveProfileAction(e) {
+  async function saveProfileAction(e) {
     e.preventDefault();
-    firebase.updateProfile(profile);
+    try {
+      !profile.address
+        ? console.log(null)
+        : await addressLookUp(profile.address).then((data) =>
```
azmainamin, line 84: wait...why do we have a await and also a .then()...

thirdeyeclub: I'll remove the async / await

PR #244, opened 2020-04-15
Feat/story 116 missions map list view

mat10tng:
This is a draft request, since this is a big PR, I figure I should open this up earlier so ppl can take a look of it 

//Edit, now open for review


src/app/dashboard/MissionsControl.jsx
```diff
@@ -45,10 +47,18 @@ const useStyles = makeStyles((theme) => ({
  *
  * @component
  */
-const MissionsPage = ({ history }) => {
+const MissionsPage = () => {
   const classes = useStyles();
   const [open, setOpen] = React.useState(false);
 
+  useFirestoreConnect(() => [
```
qdozaq, line 54: This was my main change, which moves the store logic to the root of the dashboard.
I we only connect here and let it populate the redux state we can just grab these values through the redux state instead of connecting to the db in each component.

We may still need to consider the reads that happen when an update happens. Idk if firebase optimizes that process so something to keep in mind

PR #248, opened 2020-04-16
Enhance/new model take2

mat10tng:
Ok, lots of changes due to merge in from master . So, I will list the file that are important to look at:

We defined the schema here
- src/app/model/schema.ts
My base model:
- src/app/model/BaseModel.ts
we have a few helper methods there to load, and sanitize incoming data

Model implemented on
- src/app/model/User.ts
- src/app/model/Mission.ts

Specifically, the User model now have a few methods, such as
volunteerMission, startMission and deliverMission

These methods are then use in these pages
src/app/page/Missions.jsx -> volunteerMission() ->
src/app/page/MissionAssigned.jsx -> startMission()->
src/app/page/MissionsStarted.jsx -> deliverMission()->

I have been a bit lazy here and only apply loads on the 
src/app/page/Missions.jsx 
in const mapStateToProps line 67
```
const mapStateToProps = (state, ownProps) => {
  let missions = state.firestore.ordered.missionsUnassigned;
  missions = Mission.loads(missions);
  return {
    missions: missions,
    auth: state.firebase.auth,
  };
};
```
Eventually, when update data into the firebase, we would want to use sanitize function as well. 
But since this is a big changes, I will leave it at that for now

Finnaly:
added a few new pages

https://tinyurl.com/yarmzb32
https://tinyurl.com/y8rluagd
https://tinyurl.com/yb6toj7t
https://tinyurl.com/y85jbgke


mat10tng:
https://tinyurl.com/yarmzb32
https://tinyurl.com/y8rluagd
https://tinyurl.com/yb6toj7t
https://tinyurl.com/y85jbgke

image
image

image

image

src/app/model/BaseModel.ts
```diff
@@ -0,0 +1,78 @@
...
+    return this.fs.collection(collectionName);
+  }
+
+  /**
+   * @param data: the data to load in and create a model object
+   * all fields outside of the schema will not be loaded in
+   * @returns: the object with all fields
+   */
+  load(data: unknown) {
+    return data ? merge(data, this.defaultData) : data;
```
qdozaq, line 41: Do we not want the `defaultData` returned here if there is no data?

src/app/model/BaseModel.ts
```diff
@@ -0,0 +1,78 @@
...
+    });
+  }
+
+  /**
+   * @param data: the data to load in and create a model object
+   * all fields outside of the schema will not be loaded in
+   * @returns: the object with all fields sanitized
+   */
+  sanitize(data: unknown) {
+    return data ? sanitize(data, this.defaultData) : data;
```
qdozaq, line 62: We definitely only want to return `return sanitize(...)` so we know the data is properly sanitized.
This works for `undefined`
but if the value is `false` `null` or `""` then we pass it through without checking if that was the correct type for that object.

src/app/model/Mission.ts
```diff
@@ -0,0 +1,68 @@
...
+  MissionFundedStatus,
+  MissionPayableStatus,
+  MissionDetails,
+} from "./schema";
+import BaseModel from "./BaseModel";
+
+const defaultLocation: Location = {
+  address: "",
+  lat: 0,
+  long: 0,
```
qdozaq, line 17: does this need to be `lng`? Or did you fix that?

src/app/model/User.ts
```diff
@@ -0,0 +1,107 @@
+import { UserInterface, VolunteerStatus, Location, MissionStatus } from "./schema";
+import BaseModel from "./BaseModel";
+
+const defaultLocation: Location = {
```
qdozaq, line 4: Wonder if we should just define the defaults along side the types. Then we could share them like we do the types

src/app/model/User.ts
```diff
@@ -0,0 +1,107 @@
...
+  VolunteerStatus = VolunteerStatus;
+
+  /**
+   * User volunteer for a mission
+   * @param {string} userId : user
+   * @param {string} missionId : mission that user want to volunteer for
+   */
+  async volunteerMission(userId: string, missionId: string) {
+    let collection = this.getCollection("missions");
+    let doc = await collection.doc(missionId).get();
```
qdozaq, line 37: Wrap in try/catch?

src/app/model/User.ts
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  /**
+   * User start a mission
+   * @param {string} userId - user
+   * @param {string} missionId - mission that user want to start
+   */
+  async startMission(userId: string, missionId: string) {
+    let collection = this.getCollection("missions");
+    let doc = await collection.doc(missionId).get();
```
qdozaq, line 64: same here

src/app/model/User.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  /**
+   * User deliver a mission
+   * //TOD add image
+   * @param {string} userId - user
+   * @param {string} missionId  - mission id that user delivered
+   */
+  async deliverMission(userId: string, missionId: string) {
+    let collection = this.getCollection("missions");
+    let doc = await collection.doc(missionId).get();
```
qdozaq, line 89: and here

src/app/model/schema.ts
```diff
@@ -104,16 +103,16 @@ export enum MissionType {
...
-interface FoodBoxDetails extends MissionDetails {
+export interface FoodBoxDetails extends MissionDetails {
   boxes?: Array<FoodBox>; // if multple boxes of same type grouping can occur in the UI
 }
 
 // created by the organiser.. but for MVP.0 we *hard code* a list of one type of box
-@Collection("foodboxes")
-export class FoodBox {
-  id!: string;
+// @Collection("foodboxes")
```
qdozaq, line 113: We're not using the `@Collection` tags anymore right?

src/index.tsx
```diff
@@ -22,7 +22,7 @@ const rrfConfig = {
 
 firebase.initializeApp(firebaseConfig);
 const firestore = firebase.firestore();
-fireorm.initialize(firestore);
+//fireorm.initialize(firestore);
```
qdozaq, line 25: delete?

src/app/model/Mission.ts
mat10tng: yeah, it need to lng

src/app/model/BaseModel.ts
mat10tng: no, I will remove this

src/app/model/BaseModel.ts
mat10tng: fixed

src/app/model/schema.ts
mat10tng: removed

package.json
```diff
@@ -13,6 +13,7 @@
     "@reduxjs/toolkit": "^1.3.4",
     "algolia-places-react": "^1.5.1",
     "class-validator": "^0.11.1",
+    "dataclass": "^1.2.0",
```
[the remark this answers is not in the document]
mat10tng: yes, on it

src/app/model/User.ts
mat10tng: we can wait with that for now, I do not expect defaultData to be touch from outside

src/app/model/User.ts
mat10tng: done

src/app/model/User.ts
mat10tng: done

src/index.tsx
mat10tng: done

PR #250, opened 2020-04-16
Feat/143 paypal integration

qdozaq:
This is just the Paypal component needed for stories [143](https://airtable.com/tblRjd2McKgRW9MKj/viwlRRNI0PZIxrGkc/recYoIuMMeHo45Iag?blocks=hide) and [144](https://airtable.com/tblRjd2McKgRW9MKj/viwlRRNI0PZIxrGkc/recNH5cqwqQmRIRZc?blocks=hide)

example:
image


src/app/component/PaypalCheckout/PaypalTypes.ts
```diff
@@ -0,0 +1,128 @@
+/**
```
mat10tng, line 1: should we contain the type here, or should it eventually go into the schemes folder as well?

src/app/hooks/usePaypal.js
```diff
@@ -0,0 +1,30 @@
+import { useState, useEffect } from "react";
+
+// We will eventually want to grab this from the organizer profile
+const CLIENT_ID = "sb";
+const PAYPAL_SCRIPT = `https://www.paypal.com/sdk/js?client-id=${CLIENT_ID}`;
```
mat10tng, line 5: we really need a const.js or somewhere we can put all of these config

src/app/component/PaypalCheckout/PaypalTypes.ts
qdozaq: Moving to scheme folder would be good. Especially once payments are integrated with firebase

Issue #243, opened 2020-04-15
MakeMission page doesn't use new schema


mTRx0:
The MakeMission page doesn't use the new schema to create Missions.

PR #251, opened 2020-04-16
Applied new schema to MakeMission page, added snackbar and redirect

### LINKED ISSUE #246: Creating a new mission does not show any success/error message
Route: /missions/new

Use a Success/Error snackbar (component already exists) depending on the result of mission creation to give feedback to the user whether they successfully created the mission or not.

Then redirect to My Requests.

mTRx0:
The MakeMission page now creates missions based on the new schema. If the creation was successful a success snackbar shows up (onClose the user will be redirected to /missions). Same for errors. 

I've also added a basic input validation (checks if any inputs are empty), but there's still work to do.

closes #246 #243


src/app/model/Mission.ts
```diff
@@ -63,6 +64,31 @@ class Mission extends BaseModel {
...
+   * @param {object} mission
+   * @return {string}
+   */
+  async create(mission: MissionInterface): Promise<string> {
+    const missionId = uuidV4();
+    const collection = this.getCollection(this.collectionName);
+
+    //Add mission id and sanitize data
+    mission.id = missionId;
+    const sanitizedMission = this.sanitize(mission);
```
qdozaq, line 80: Decided this in a separate convo, but we're gonna use `this.load`(maybe needs a new name?) instead for creates and save `sanitize` for updates. `load` will default any missing data on the object as well as sanitize it

src/app/model/User.ts
```diff
@@ -27,6 +27,34 @@ const defaultUserData: UserInterface = {
...
+    } catch (error) {
+      //TODO show error message to user
+      throw error;
+    }
+
+    if (doc.empty) {
+      throw Error(`This user: ${displayName} does not exist`);
+    }
+
+    doc.forEach((data) => {
```
azmainamin, line 49: This is for my own understanding: why are we checking for each item in doc?

src/app/model/Mission.ts
mTRx0: I've changed this in the new PR #259, could you have a look over there? Honestly still not 100% sure how to correctly use the `load` and `sanitize` functions.

src/app/model/User.ts
```diff
@@ -27,6 +27,34 @@ const defaultUserData: UserInterface = {
...
+      if (data === undefined) {
+        throw Error(`no data for this user: ${displayName}`);
+      }
+    });
+
+    return doc.docs[0].id;
```
azmainamin, line 55: I would propose returning the whole user object. That way this method is more useful.

src/app/page/MakeMission/MakeMission.jsx
```diff
@@ -6,73 +6,163 @@ import MissionForm from "./MissionForm";
...
+          throw Error("image upload error");
+        });
+      return await uploadTask;
     } else {
-      saveMissions(val);
+      return "";
+    }
+  }
+
+  async function validateInput(payload) {
```
azmainamin, line 105: Minor: Rename to validateAndSaveMission() ;

src/app/model/Mission.ts
azmainamin: @mTRx0  yep! That's how I understand it. I am learning about TS on the go.

qdozaq: Sorry if that was confusing. Yeh use `load` for creating new objects and fetching objects from the db and `sanitize` for updates.

package.json
```diff
@@ -11,6 +11,7 @@
     "@material-ui/lab": "^4.0.0-alpha.49",
     "@material-ui/pickers": "^3.2.10",
     "@reduxjs/toolkit": "^1.3.4",
+    "@types/uuid": "^7.0.2",
```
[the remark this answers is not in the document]
mTRx0: I've only added uuid type definitions for ts. Needed for generating a mission id in the createMission function (Maybe it should be generated outside of it?)

src/app/model/User.ts
mTRx0: Oh, that's a mistake. Just copied that part from the existing functions and didn't think about (or remove) it.

src/app/model/User.ts
mTRx0: Only the first user object or the whole array? Right now the displayNames are not exclusive to each user so there could be multiple results...

src/app/model/Mission.ts
mTRx0: Alright, got it. Thanks!

package.json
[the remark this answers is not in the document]
azmainamin: Gotcha! This is good.

src/app/model/User.ts
azmainamin: Ah. If that's the case, how do we decide which Id to return? Just the first one?

mTRx0: I don't have any better idea in mind.

azmainamin: If this works, this works! I have no problem.

mTRx0: The only problem is the user input. I guess the problem will be resolved as soon as the user search component (PR #252) is merged as this should return an user id instead of the displayName.

azmainamin: That's awesome! We can refactor this later then. Let's get this PR in to master. I will start my work after that :P

PR #252, opened 2020-04-16
feat: add resuable user search component

AquiburRahmanKhan:
This PR is related to https://airtable.com/tblRjd2McKgRW9MKj/viwlRRNI0PZIxrGkc/rec00kGce0l6ZBJUr?blocks=hide


qdozaq:
Some changes went to update the models. You'll need to fix those conflicts

src/app/component/SearchUser.jsx
```diff
@@ -0,0 +1,92 @@
...
+
+const useStyles = makeStyles((theme) => ({
+  root: {
+    paddingBottom: theme.spacing(2),
+  },
+  avatar: {
+    marginRight: theme.spacing(1),
+  },
+  textField: {
+    paddingRight: "9px!important",
```
azmainamin, line 18: Can we not use !important ? This might cause potential UI bugs in the future.

src/app/component/SearchUser.jsx
```diff
@@ -0,0 +1,92 @@
...
+  const [open, setOpen] = useState(false);
+
+  const handleInputChange = async (event, value, reason) => {
+    if (value.length > 1) {
+      setLoading(true);
+
+      const users = await Users.usersMatchingLabel(value, 10);
+      setUsers(users);
+
+      if (reason == "input") {
```
azmainamin, line 36: '==='

src/app/component/SearchUser.jsx
```diff
@@ -0,0 +1,92 @@
...
+
+  const [users, setUsers] = useState([]);
+  const [loading, setLoading] = useState(false);
+  const [open, setOpen] = useState(false);
+
+  const handleInputChange = async (event, value, reason) => {
+    if (value.length > 1) {
+      setLoading(true);
+
+      const users = await Users.usersMatchingLabel(value, 10);
```
azmainamin, line 33: Can we use a named variable for the magic number (10) ?

src/app/component/SearchUser.jsx
```diff
@@ -0,0 +1,92 @@
...
+      options={users}
+      open={open}
+      filterOptions={(options) => options}
+      getOptionLabel={(option) => option.displayName}
+      onClose={() => {
+        setOpen(false);
+      }}
+      loading={loading}
+      onChange={handleChange}
+      onInputChange={_.debounce(
```
azmainamin, line 62: Love the debounce!

src/app/model/schema.ts
```diff
@@ -48,7 +48,7 @@ export class User {
   FIXME: need to ensure this is synced from firebase.auth ph number
   FIXME: where do we assert phone number formatting?
   FIXME: currently, always null */
-  phone?: number;
+  phoneNumber?: number;
```
azmainamin, line 51: +++

src/app/page/TypeAheadDemo/TypeAheadDemo.jsx
```diff
@@ -0,0 +1,24 @@
+import React, { useState } from "react";
```
azmainamin, line 1: Is this a temp component?

src/app/component/SearchUser.jsx
jwu910: +1 for this. Strongly believe !important should never be used. If our styles are set up correctly, we should not need to force specificity

AquiburRahmanKhan: Understood, will remove important!

src/app/component/SearchUser.jsx
AquiburRahmanKhan: Will do

src/app/page/TypeAheadDemo/TypeAheadDemo.jsx
AquiburRahmanKhan: Yes, this is just so you can see the userSearch Component in action, this can be thrown away when you guys actually use the userSearch Component somewhere

utunga:
@AquiburRKhan / @mat10tng so, I 
- pulled latest from master and rebased this branch on that
- made those schema changes we discussed, Tuan
- got it working again - search logic works according to the new architecture, and search happens on the server but unfortunately due to[ limitations of firebase ](https://stackoverflow.com/questions/46568142/google-firestore-query-on-substring-of-a-property-value-text-search ) it's *still* impossible to actually get this to deliver the desired behavior.

So.. RIght now you have to type the first name including with Caps for it to work. 

That said I would request that we merge this in anyway.
- The UI component itself works as intended
- By merging you get those schema changes to master, which is the main reason
- I will revisit this again later to make the type ahead work nicer either by switching back to having the complete list of users on the client or by installing free text search engine such Agolia or similar on the server

utunga:
@AquiburRKhan after the rebase I had to do a `git push -f` after the rebase ...

azmainamin:
@AquiburRahmanKhan @utunga anything left for this?

azmainamin:
@AquiburRahmanKhan i see tests failing

.gitignore
```diff
@@ -26,3 +26,4 @@ yarn-error.log*
 .vscode/# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
 
 .env
+.vscode
```
jwu910, line 29: This looks like it's added twice.

src/app/page/TypeAheadDemo/TypeAheadDemo.jsx
jwu910: Thanks for sharing this demonstration! However, I don't think we should add test components into the source code like this.

I think sharing some sample code is enough to demonstrate usage, or taking a moment to see if a storybook component can be set up.

utunga:
Yeah there's stuff left for this. I've got a new idea of what we need to do to make this 'mergeable'. Basically forget the full text search and just get a full list of volunteers client side (again). its honestly fine for MVP we may only have a handful of volunteers anyway!!

src/app/page/TypeAheadDemo/TypeAheadDemo.jsx
AquiburRahmanKhan: Understood, will throw it away then and share a sample code

AquiburRahmanKhan:
> @AquiburRahmanKhan @utunga anything left for this?

Yes from my side, I wasn't able to see to the comments left by you and jwu910 , I'll get to them today and then let you know. Sorry for the delay

src/app/page/TypeAheadDemo/TypeAheadDemo.jsx
AquiburRahmanKhan: @jwu910 where do I share the sample usage code for this ?

AquiburRahmanKhan:
@azmainamin I addressed all the feedback comments and also removed the typeaheadDemo component, please review it again and I think it is ready to get merged from my side, but Miles has some work left on the backend maybe.

utunga:
Yeah I'd like to simplify the back end a bit before we merge. What we have
now makes no sense.

Miles Thompson
NZ: +64 2153 9843 | E: utunga@gmail.com | W: http://goodbright.nz

utunga:
@AquiburRahmanKhan with my apologies can I retract my previous comment. I think the purpose of this component is the front end and its reasonable to consider the logic that provides the list of users to this component to be a separate card... in some cases It will be a list of active volunteers, in other cases a list of recipients etc. So I would like to suggest we just get this one merged sooner rather than later so that we don't hold things up. 

I stupidly put some schema changes in here too that @mat10tng had requested so, yeah.

Also Im taking a day off so won't be around to finish that back end method right now.

src/app/page/TypeAheadDemo/TypeAheadDemo.jsx
jwu910: Hey @AquiburRKhan You always share as a comment or in this PR description.

You can also consider looking at helping build out storybook components so other devs can see samples. The Storybook library is pretty small right now, and i'm sure additions would be appreciated.

Issue #247, opened 2020-04-16
finish refactor of repository architecture 


utunga:
Tuan to finish off one example ? and then maybe someone else can jump on and finish applying this across all the code

mat10tng:
The first version are now in.

qdozaq:
@mat10tng can we get a list of the remaining models that need to be done? then @azmainamin and I could split them up

mat10tng:
Yes!
models:
Organization
Foodbox
Page:
MakeMission -> I would like to change this to MissionCreate and use the Mission model to load and then sanitize the data to be save in the backend with User.createMission
We also have to change some of the field that are not in use in this view, but in organizer view
MakeRequest-> remove this, we use MakeMisssion now
MissionDetails -> use the model to load, and we need to sanitize it before send back to database
OrganizerSignup -> ??? maybe MVP1 but same with MissionCreate
Signup -> making sure use created with full information, same with MissionCreate
Status -> not sure if we using this right now, maybe remote from route, but keep the page




Organizations ( though admittedly, that

Eihcir0:
I can take signup on this

azmainamin:
@Eihcir0  Signup is yours.
@mTRx0  MakeMission is yours. I think removing MakeRequest falls under this as well?

I will work on

qdozaq:
I'll do OrganizerSignup? is that needed?

mTRx0:
> @mTRx0 MakeMission is yours. I think removing MakeRequest falls under this as well?

Yeah I can do that aswell.

azmainamin:
@mTRx0 I am reviewing your PR. Can you make any other changes in a different PR? Thank!

mTRx0:
sure :)

utunga:
@qdozaq  should we close this issue or is it still outstanding?

qdozaq:
Yeh we can close this

PR #259, opened 2020-04-16
Enhancement/repository refactor

mTRx0:
continues PR #251 with changes from https://github.com/factn/resilience-app/issues/247#issuecomment-614764357


azmainamin:
Will start reviewing this after PR #251 is merged.

src/app/model/User.ts
```diff
@@ -28,10 +35,34 @@ class User extends BaseModel {
...
+  async createMission(mission: MissionInterface): Promise<string> {
+    const missionId = uuidV4(); //generate mission id
+    const collection = this.getCollection("missions");
+
+    //Add mission id to mission object
+    mission.id = missionId;
+
+    //save mission in firestore
+    try {
+      await collection.doc(missionId).set(mission);
```
qdozaq, line 52: so yeh here is where we would want to use `load`

qdozaq: Also any particular reason this was moved to `User` instead of `Mission`?

mTRx0: > so yeh here is where we would want to use `load`
 
The `load` call is outside of this function in MissionCreate.jsx (If that's what you mean). Moved it because @mat10tng mentioned he wants the data to be sent to backend in User.createMission in #247.

mat10tng: for the sake of sanity when create object, we should do an extra load here, just in case ppl mutate the object outside without our knowing

mTRx0: Yeah I guess that makes a lot more sense!

PR #274, opened 2020-04-18
Dashboard data optimization. Make MissionDetailsCard Stateless

qdozaq:
We already have all the missions loaded in the dashboard. no reason to fetch one again when we open a details card.
Pr sends mission object directly to component now instead of grabbing it from firestore again

### DIFF
--- a/src/app/dashboard/Missions/DashboardMissions.jsx
+++ b/src/app/dashboard/Missions/DashboardMissions.jsx
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React, { useState, useEffect } from "react";
 import { makeStyles } from "@material-ui/core/styles";
 import Grid from "@material-ui/core/Grid";
 import { Button } from "../../component";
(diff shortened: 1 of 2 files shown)


src/app/dashboard/Missions/DashboardMissions.jsx
```diff
@@ -133,15 +133,24 @@ const PageButtons = ({ classes, pageView, setPageView }) => {
 
 const DashboardMissions = ({ inDone, inPlanning, inProgress, inProposed }) => {
   const classes = useStyles();
-  const all = { inProposed, inPlanning, inProgress, inDone };
 
   const [pageView, setPageView] = useState(pageViews.list);
-  let filtered = inProposed;
+  const [selectedMission, setSelectedMission] = useState(null);
 
+  const all = { inProposed, inPlanning, inProgress, inDone };
```
jwu910, line 140: Would it be more clear to pass this object in named as missions from the parent, and access or destructure this similar to  props.missions.inProgress ?

src/app/dashboard/Missions/DashboardMissions.jsx
```diff
@@ -133,15 +133,24 @@ const PageButtons = ({ classes, pageView, setPageView }) => {
...
   const viewFromUrl = _.getQueryParam("view");
-  filtered = all[viewFromUrl];
+  const filtered = all[viewFromUrl] || inProposed;
+
+  const selectedMissionId = _.getQueryParam("missionId");
```
jwu910, line 144: I’ll have to take a look at our router set up more in-depth later, or not from my phone,  but react router should at least provide this to us in props with something like props.match.params.missionId

qdozaq: Yeh I think our use of react router could def be better

PR #275, opened 2020-04-18
story118/feat/add funded status select field

mat10tng:


firebase-debug.log
```diff
@@ -0,0 +1,17 @@
+[debug] [2020-04-18T15:55:13.624Z] ----------------------------------------------------------------------
```
jwu910, line 1: I think we should just add this to the gitignore

src/app/dashboard/Missions/ListComponents.jsx
```diff
@@ -1,52 +1,138 @@
...
-function MissionName({ onShowDetails, title, type }) {
+const UnasignedStatusAction = ({ missionId }) => {
+  const classes = useStyles();
+  const handleChange = (event) => {
+    event.preventDefault();
+    Mission.update(missionId, {
+      fundedStatus: event.target.value,
+      status: Mission.Status.tentative,
+    });
+  };
```
jwu910, line 41: If we're directly updating the mission status via one select change handler, and then the organizer accidentally closes the page, there may be some unintended side effects.

Do we need to consider adding in a confirmation or button to submit a change?

src/app/dashboard/Missions/MissionDetails.jsx
```diff
@@ -93,7 +93,7 @@ const MissionDetailsCard = ({ mission, setselectedMissionId }) => {
     <Grid item xs={3}>
       <Paper className={classes.root} elevation={0}>
         <Grid container direction="row-reverse">
-          <Button onClick={() => setselectedMissionId(null)} variant="text">
+          <Button onClick={() => setSelectedMissionId(null)} variant="text">
```
jwu910: :ok_hand:

src/app/dashboard/Missions/MissionsListView.jsx
```diff
@@ -74,7 +71,7 @@ const config = [
   {
     format: (mission) => ({
       time: mission.deliveryWindow,
-      location: mission.deloveryLocation,
+      location: mission.deliveryLocation,
```
jwu910, line 74: Nice catch

src/app/dashboard/Missions/MissionsListView.jsx
```diff
@@ -98,6 +99,7 @@ const MissionsListView = ({ missions, view }) => {
   const theme = useTheme();
   const innerTheme = getMuiTheme(theme);
   const history = useHistory();
+  console.log(missions);
```
jwu910, line 102: Can we remove unnecessary console logs?

src/app/model/Mission.ts
```diff
@@ -50,8 +50,8 @@ const defaultMissionData: MissionInterface = {
   recipientName: "No Recipient Name",
   recipientPhoneNumber: "",
   recipientId: "No Recipient Id", // reference?
-  created: new Date(), // time stamp
-  lastUpdated: new Date(), // time stamp
+  created: "", // time stamp
+  lastUpdated: "", // time stamp
```
jwu910, line 54: Haven't had a chance to fully review the models, will changing the model defaults to an empty string affect other components asking for this?

src/index.css
```diff
@@ -6,6 +6,10 @@ body {
   -moz-osx-font-smoothing: grayscale;
 }
 
+* {
+  margin: 0;
+}
```
jwu910, line 11: :open_mouth: What margins being removed with this? Would this override many of the MUI pre-defined margins?

src/app/model/Mission.ts
mat10tng: it just not work with Date...that is why we opted for string

src/index.css
mat10tng: no, mui has its own def, I think we should just have a reset.css somewhere tbh, but I opted in for this only because of how the designer expect the header to work vs how the header margin actually work..

firebase-debug.log
mat10tng: remove and add to gitignore

src/app/dashboard/Missions/ListComponents.jsx
mat10tng: yes, I have suggest another flow with modal, but we are going to wait for the next mvp for that

jwu910: Okay cool, just wanted to keep that on the radar

PR #284, opened 2020-04-19
Signup hookup

Eihcir0:
- Updates User schema -- adds email, description, and availability fields
- Adds saveNewUser method to User model
- Updates front end logic to save new user


src/app/page/Signup/SignupScene.jsx
```diff
@@ -1,35 +1,18 @@
...
 import PropTypes from "prop-types";
-import useForm from "../../hooks/useForm";
 
+import useForm from "../../hooks/useForm";
 import CallToActionPage from "./CallToAction";
 import UserProfilePage from "./UserProfile";
 import ConnectSocialMediaPage from "./ConnectSocialMedia";
 import PhoneAuthPage from "./PhoneAuth";
 import SignupSuccessPage from "./SignupSuccess";
-
```
Eihcir0, line 11: note: Moved to `./helpers.js`

src/app/page/Signup/SignupSuccess/SignupSuccess.jsx
```diff
@@ -16,10 +15,6 @@ function SignupSuccess({ handleButtonClick }) {
   return (
     <Page>
       <Container classes={{ root: classes.root }}>
-        <SuccessSnackbar
```
Eihcir0, line 19: note: Moved to top level SignupScene

src/app/model/User.ts
```diff
@@ -35,6 +38,20 @@ const defaultUserData: UserInterface = {
 class User extends BaseModel {
   VolunteerStatus = VolunteerStatus;
 
+  async saveNewUser(data: UserInterface) {
+    const id = uuidV4(); //generate User id  TODO: Setup firebase tables to autogenerate ids
```
jwu910, line 42: If i recall, user IDs generated through firebase were just auto incrementing numbers. @qdozaq, Do you know if that's true?

src/app/model/schema.ts
```diff
@@ -44,13 +44,16 @@ export class UserInterface {
   FIXME: need to ensure this is synced from firebase.auth ph number
   FIXME: where do we assert phone number formatting?
   FIXME: currently, always null */
-  phone?: number;
+  phone!: string;
```
jwu910, line 47: Have we checked if this type change affects any other components consuming this information?

src/app/page/Signup/SignupScene.jsx
```diff
@@ -38,6 +21,41 @@ const convertFullName = (fullName) => {
...
+      email: values.email || "",
+      isOrganizer: false,
+      isVolunteer: true,
+      location: values.location && convertLocation(values.location),
+      organizerDetails: {},
+      phone: values.phone || "",
+      photoUrl: "",
+      volunteerDetails: {
+        availability: values.availability || "",
+        hasTransportation: Boolean(values.hasTransportation),
```
jwu910, line 40: hasTransportation is already boolean isnt it?

src/app/page/Signup/helpers.js
```diff
@@ -0,0 +1,33 @@
...
+  return [firstName, lastName];
+};
+
+/**
+ * Takes a location object provided by the Angolia location selector
+ * and converts it to fit the model field
+ * @param {object} location { address, countryCode, county, lat, long }
+ * @function
+ */
+export const convertLocation = (location) => {
```
jwu910, line 29: Possible to rename this to `normalizeLocation` for semantic understanding?

src/app/page/Signup/helpers.js
```diff
@@ -0,0 +1,33 @@
+/**
```
jwu910, line 1: Commenting for this entire file:

Should these functions be shared across the app? If there are opportunities to not rewrite these functions elsewhere, maybe we can move these to a shared util directory  not specific to Signup

src/app/page/Signup/SignupScene.jsx
Eihcir0: it could be undefined

src/app/model/User.ts
mat10tng: image

mat10tng: Ok, about user ID, once a user sign up through auth, they got an id, this is stored inside firebase.auth.uid

This uid should be used as the key 

This can be obtain by during sigun up step, you can get the result.user -> set this inside the users collection
image
Link for ref:  https://firebase.google.com/docs/auth/web/phone-auth

Eihcir0: @mat10tng 
Ok, it is easy to get the uid from the phone auth which I can use later when saving the profile.  

But I am not sure about the next step after PhoneAuth in the signup process -- **Connect social media**.  As of now we are 'authenticating' again and then we get back an entirely different uid from the one we get from phone auth.

I think I will have to research more about [linking multiple auths ](https://firebase.google.com/docs/auth/web/account-linking)-- or depening on the rush of this pr maybe we just ignore it for now?  Especially since @qdozaq said we're doing a complete overhaul of firebase...

One more thing, I noticed the format of the uuid provided from the auth:
`IpENTHiC3bfPB6h6CNsWZ2izC123
`
 look different from the uuid generated by that `uuidv4():
`1e8d498f-27b3-4972-89e3-ee07a31ea183
`

src/app/page/Signup/helpers.js
Eihcir0: ok

src/app/model/User.ts
mat10tng: no worry, the auth.uid is always the same.

Eihcir0: I'm getting 2 different uuid's -- one for the phone auth and one for the social media auth.
from phone auth response:
Screen Shot 2020-04-19 at 4 34 12 PM

from google auth response:
Screen Shot 2020-04-19 at 4 34 56 PM

This is using the auth functions we had in place before I started this.  I think we need to somehow pass the uid we get from the phone auth to the social media auth call, but I would need to look more into that.

src/app/page/Signup/helpers.js
Eihcir0: moved to `utils/`

src/app/model/User.ts
Eihcir0: After conversation with @mat10tng we've decided to not focus on solving this issue right now so we can work on other things for mvp0.

What we have now will authenticate phone and connect that with the user profile information filled out by the new user.  It will also authenticate the social media if chosen and retain the email/displayName from that. 

We've created [this ticket](https://github.com/factn/resilience-app/issues/290) to track open items / edge cases around the Signup flow.

PR #294, opened 2020-04-21
Feat/story 101 homepage with request and signup

amanchauhan11:
Story 101 homepage iteration https://airtable.com/tblRjd2McKgRW9MKj/viwlRRNI0PZIxrGkc/reckphMmrtbyDUwXm?blocks=hide

Mock up:
https://www.figma.com/file/v5HdxDsvO2NzMuZU6CArzF/Resilience-App?node-id=1810%3A21

I have created a new Appbar `PublicAppbar` and renamed earlier one to `DefaultAppbar`. `PublicAppbar` can be used in `About us` page as well. Modified `Home.js` to per new design. At the moment I have linked the `Donate` button to sign up page as there isn't a separate page for donations.


src/app/page/Home/Home.js
```diff
@@ -80,62 +144,100 @@ const HomePage = ({ auth, history }) => {
...
-              </Grid>
-            </Grid>
-          </Card>
-        </Grid>
+            </Box>
+          </Box>
+          <Box className={classes.RequestByPhoneBg}>
+            <Card className={classes.PhoneCard}>
+              <img src={PhoneIcon} alt="call"></img>
+              <Text>To request food by phone, call</Text>&nbsp;<PhoneNo>555-555-5555.</PhoneNo>
```
qdozaq, line 232: Can we use an anchor tag here for the number instead?
you can add `tel:` to href to make auto call that number
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a

src/app/page/Home/Home.style.js
```diff
@@ -33,12 +33,44 @@ export const StyledHomeButton = styled(Button)`
...
+
+export const Text = styled.span`
+  font-family: Open Sans, Ariomo, Arial, sans-serif;
   font-size: 16px;
-  line-height: 20px;
-  color: #150e60;
-  text-decoration: none;
+  line-height: 22px;
+  font-weight: ${props => props.bold ? "bold" : "normal"};
+  color: ${props => props.color ? props.color : props.theme.color.black};
```
qdozaq, line 54: This is fine, but if you want to condense in the future you could also say
```ts
=> props.color || props.theme.color.black
```

src/app/page/Home/Home.js
```diff
@@ -1,16 +1,24 @@
...
+import HomeImage5 from "../../../img/HomeImage5.svg";
 import SplashImage1 from "../../../img/SplashImage1.png";
-import { Body1, Button, H1, H2, H3 } from "../../component";
-import { Card, Page } from "../../layout";
+import Emoticon from "../../../img/ic_smiley_emoticon.svg";
+import PhoneIcon from "../../../img/ic_phone.svg";
+
 import VolunteerHome from "./VolunteerHome";
 
 const useStyles = makeStyles((theme) => ({
```
qdozaq, line 24: Could we move this to Home.style.js as well?

azmainamin:
Build failed as well due to failed test.

src/app/page/Home/Home.js
```diff
@@ -56,6 +59,67 @@ const useStyles = makeStyles((theme) => ({
...
+    display: "flex"
+  },
+  Emoticon: {
+    height: "36px",
+    width: "36px",
+    marginLeft: theme.spacing(1),
+    verticalAlign: "middle"
+  },
+  PurpleBg: {
+    backgroundColor: theme.color.deepPurple
```
jwu910, line 102: Use of deepPurple doesn't seem to exist

mat10tng: its actually red now, I dont know, the theme keep changing from the designer, lets add deepPurple back ....
but it is red.

jwu910: Can we use less specific variables for these? like backgroundPrimary or backgroundLight or something that isn't tied to a color? So if Design decides that these colors change, we just update that variable at the theme level, and changes like this wouldn't be affected

amanchauhan11:
I think this page has also been completed by @priyankpat here [.](https://github.com/factn/resilience-app/pull/307) So I guess we are keeping that one?

src/app/page/Home/Home.js
mat10tng: I would like that as well, but it is the design in Figma that show up as color red/deepPurple instead. We should ask this question to the designer

priyankpat: Can we stick to the colors provided by [material-ui library](https://material-ui.com/customization/color/)?

qdozaq:
Talked to @amanchauhan11 and looks like we're going to go with the #307 implementation of the homepage. Closing.

PR #295, opened 2020-04-21
Added snackbar to mission list

### LINKED ISSUE #271: Volunteering for a mission from the mission list view(/missions) gives me no message whether the request was successful or not
We should add a snack bar stating whether the action (volunteering for a mission) was successful or not.

mTRx0:
closes #271

### DIFF
--- a/src/app/page/Missions/Missions.jsx
+++ b/src/app/page/Missions/Missions.jsx
@@ -19,10 +19,13 @@ import UserPhoneUnverifiedPopup from "../../component/UserPhoneUnverifiedPopup";
  * @param {object} props.history - Object obtained from React Router
  */
 const MissionsPage = ({ auth, history, missions }) => {
+  const initSnackbarState = {
+    open: false,
+    type: "error",
+    message: "",
+  };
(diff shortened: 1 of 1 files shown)


src/app/page/Missions/Missions.jsx
```diff
@@ -19,14 +20,27 @@ import UserPhoneUnverifiedPopup from "../../component/UserPhoneUnverifiedPopup";
  */
 const MissionsPage = ({ auth, history, missions }) => {
   const [popupOpen, setPopupOpen] = useState(false);
+  const [snackbarOpen, setSnackbarOpen] = useState(false);
```
azmainamin, line 23: We can probably combine the state properties for the snackbar into an object:
`const initialSnackbarState = {open: false, type: 'error', message: ""}
const [snackbarState, setState] = useState(initialSnackbarState);
...
setState({open:true, type: error, message: "Message"});
`

mTRx0: Yeah that makes sense. Will correct it.

PR #304, opened 2020-04-22
Feature/story 109 Location update

priyankpat:
[Story 109](https://airtable.com/tblRjd2McKgRW9MKj/viwlRRNI0PZIxrGkc/recLNIOL5Lpg4odOB?blocks=hide)
[Figma](https://www.figma.com/file/v5HdxDsvO2NzMuZU6CArzF/MutualAid.world-Designs-in-progress?node-id=1565%3A769)

**Required:** `REACT_APP_MAPS_ACCESS_TOKEN` in .env.development file

Gave the location google places autocomplete in order to standardize the address format when stored in the db.
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,12 @@ npm ci
 
 Copy the file pinned to the side of #resilience-app channel to `.env.development`in the git root of this repo.
 
+- Example of .env.development content
+```
+REACT_APP_MB_ACCESS_TOKEN=
+REACT_APP_MAPS_ACCESS_TOKEN=
+```
+
 - Start the local server
 
 ```
--- a/src/app/hooks/useScript.js
+++ b/src/app/hooks/useScript.js
@@ -1,0 +1,55 @@
+import { useEffect, useState } from "react";
+
+const cachedScripts = [];
+
+export default function useScript(src) {
+  const [state, setState] = useState({
+    loaded: false,
+    error: false,
+  });
+
+  useEffect(() => {
+    if (cachedScripts.includes(src)) {
+      setState({
+        loaded: true,
+        error: false,
+      });
+    } else {
+      cachedScripts.push(src);
+
+      let script = document.createElement("script");
+      script.src = src;
+      script.async = true;
+
+      const onScriptLoad = () => {
+        setState({
+          loaded: true,
+          error: false,
+        });
+      };
+
+      const onScriptError = () => {
+        const index = cachedScripts.indexOf(src);
+        if (index >= 0) cachedScripts.splice(index, 1);
+        script.remove();
+
+        setState({
+          loaded: true,
+          error: true,
+        });
+      };
+
+      script.addEventListener("load", onScriptLoad);
+      script.addEventListener("error", onScriptError);
+
+      document.body.appendChild(script);
+
+      return () => {
+        script.removeEventListener("load", onScriptLoad);
+        script.removeEventListener("error", onScriptError);
+      };
+    }
+  }, [src]);
+
+  return [state.loaded, state.error];
+}
--- a/src/app/page/UserProfile/ProfileImage.jsx
+++ b/src/app/page/UserProfile/ProfileImage.jsx
@@ -69,7 +69,7 @@ const ProfileImage = ({ classes, profile, setProfile }) => {
   ) : (
     <Grid item>
       <Avatar className={classes.profileImage} src={profilePhoto} alt={displayName} />
-      <Link style={{ textDecoration: "none" }} onClick={setEdit}>
+      <Link to="" style={{ textDecoration: "none" }} onClick={setEdit}>
         <h4>Change Picture</h4>
       </Link>
     </Grid>
--- a/src/app/page/UserProfile/ProfileImage.jsx
+++ b/src/app/page/UserProfile/ProfileImage.jsx
@@ -68,7 +68,7 @@ const ProfileImage = ({ classes, profile, setProfile }) => {
   ) : (
     <Grid item>
       <Avatar className={classes.profileImage} src={profilePhoto} alt={displayName} />
-      <Button className={classes.profileEditButton} onClick={setEdit}>
+      <Button color="primary" variant="text" onClick={setEdit}>
         <h4>Change Picture</h4>
       </Button>
     </Grid>
--- a/src/app/page/UserProfile/ProfileImage.jsx
+++ b/src/app/page/UserProfile/ProfileImage.jsx
@@ -3,7 +3,6 @@ import _ from "lodash";
 import React, { useState } from "react";
 import { useSelector } from "react-redux";
 import { useFirebase } from "react-redux-firebase";
-import { Link } from "react-router-dom";
 
 import { Button } from "../../component";
 import PhotoSelection from "./photoselection.jsx";
@@ -69,9 +68,9 @@ const ProfileImage = ({ classes, profile, setProfile }) => {
   ) : (
     <Grid item>
       <Avatar className={classes.profileImage} src={profilePhoto} alt={displayName} />
-      <Link to="" style={{ textDecoration: "none" }} onClick={setEdit}>
+      <Button className={classes.profileEditButton} onClick={setEdit}>
         <h4>Change Picture</h4>
-      </Link>
+      </Button>
     </Grid>
   );
 };
--- a/src/app/page/UserProfile/UserOverview.jsx
+++ b/src/app/page/UserProfile/UserOverview.jsx
@@ -58,10 +58,9 @@ const UserStatus = ({ profile, setProfile, view }) => {
       type: ["(cities)"],
     };
     const addrElem = document.getElementById("address");
-    const googleRef = window.google;
 
-    if (googleRef !== undefined) {
-      autocomplete = new googleRef.maps.places.Autocomplete(addrElem, options);
+    if (window.google !== undefined) {
+      autocomplete = new window.google.maps.places.Autocomplete(addrElem, options);
       autocomplete.setFields(["address_components", "formatted_address"]);
       autocomplete.addListener("place_changed", handleAddressSelect);
     }
--- a/src/app/page/UserProfile/UserOverview.jsx
+++ b/src/app/page/UserProfile/UserOverview.jsx
@@ -59,7 +59,7 @@ const UserStatus = ({ profile, setProfile, view }) => {
     };
     const addrElem = document.getElementById("address");
 
-    if (window.google !== undefined) {
+    if (window.google) {
       autocomplete = new window.google.maps.places.Autocomplete(addrElem, options);
       autocomplete.setFields(["address_components", "formatted_address"]);
       autocomplete.addListener("place_changed", handleAddressSelect);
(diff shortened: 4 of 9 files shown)


src/app/page/UserProfile/UserOverview.jsx
```diff
@@ -29,8 +33,33 @@ const UserStatus = ({ profile, setProfile, view }) => {
...
+    }
+  };
+
+  const handleScriptLoad = () => {
+    const options = {
+      type: ["(cities)"],
+    };
+    const addrElem = document.getElementById("address");
+
+    autocomplete = new google.maps.places.Autocomplete(addrElem, options);
```
qdozaq, line 52: Can we make it clear that this is coming from the `window` object? Also should check that this script has loaded correctly. I believe `onError` is a required prop for react-load-script module you're using here.

Something to look at later (maybe i'll make a ticket for it) is we could re purpose the [usePaypal](https://github.com/factn/resilience-app/blob/master/src/app/hooks/usePaypal.js) hook to be a general `useScript` hook for situations like this. I think hooks are better for this async behavior than having it in a react element

priyankpat: I can refactor it to use hooks for this situation and maybe generalize it for any scripts.

aamin21: Hooks would be nice!

priyankpat: There is not a **useScript** hook for generic script loading.

src/app/page/UserProfile/ProfileImage.jsx
```diff
@@ -69,7 +69,7 @@ const ProfileImage = ({ classes, profile, setProfile }) => {
   ) : (
     <Grid item>
       <Avatar className={classes.profileImage} src={profilePhoto} alt={displayName} />
-      <Link style={{ textDecoration: "none" }} onClick={setEdit}>
+      <Link to="" style={{ textDecoration: "none" }} onClick={setEdit}>
```
jwu910, line 72: Link's an element intended for router navigation. We're gonna come across some issues not using this as intended either around some required Link props or accessibility issues with this having a callback rather than initiating navigation.

src/app/page/UserProfile/UserOverview.jsx
```diff
@@ -51,7 +85,7 @@ const UserStatus = ({ profile, setProfile, view }) => {
       </Grid>
       <Grid item container spacing={1} direction="column">
         <Grid item container>
-          <H5>Address</H5>
+          <H5>Location</H5>
```
jwu910, line 88: If we're changing Address -> Location, we should probably update our variables to match to avoid confusion.

src/app/page/UserProfile/UserOverview.jsx
```diff
@@ -48,18 +50,21 @@ const UserStatus = ({ profile, setProfile, view }) => {
       type: ["(cities)"],
     };
     const addrElem = document.getElementById("address");
+    const googleRef = !!window && !!window.google ? window.google : undefined;
```
jwu910, line 53: What's going on here?

src/app/page/UserProfile/ProfileImage.jsx
qdozaq: Yup, you can use the default `Button` from `material-ui/core` instead of the one from components.
See the docs to make it look like you want https://material-ui.com/components/buttons/#button

src/app/hooks/useScript.js
```diff
@@ -0,0 +1,55 @@
+import { useEffect, useState } from "react";
+
+const cachedScripts = [];
+
+export default function useScript(src) {
```
qdozaq, line 5: 👍

src/app/page/UserProfile/UserOverview.jsx
qdozaq: These should get auto treated as booleans since they're being used to determine the ternary. no need for the double negation.

Also since this is happening after the component has mounted it's safe to assume the window object is already available. Could just say 
```js
window.google ? window.google : undefined;
//or really just
ref = window.google;
// Given that you're only checking for undefined it'll already be undefined if not present
```

src/app/page/UserProfile/UserOverview.jsx
```diff
@@ -29,6 +35,34 @@ const UserStatus = ({ profile, setProfile, view }) => {
...
+
+    if (googleRef !== undefined) {
+      autocomplete = new googleRef.maps.places.Autocomplete(addrElem, options);
+      autocomplete.setFields(["address_components", "formatted_address"]);
+      autocomplete.addListener("place_changed", handleAddressSelect);
+    }
+  };
+
+  if (loaded && !error) {
+    handleScriptLoad();
```
qdozaq, line 63: Should probably do this all inside a `useEffect` so we can watch when loaded and error change as well remove that listener on unmount

src/app/page/UserProfile/UserOverview.jsx
priyankpat: I simply changed the label to reflect the design in Figma. We can keep Address instead of Location. I think it sounds more appropriate in a user profile.

src/app/page/UserProfile/UserOverview.jsx
priyankpat: Addressed in 1ea575f

src/app/page/UserProfile/UserOverview.jsx
priyankpat: Addressed in 1ea575f

src/app/page/UserProfile/ProfileImage.jsx
priyankpat: Addressed in e552888

src/app/page/UserProfile/UserOverview.jsx
jwu910: Not a deal breaker, for code review, but just for code clarity, I think this is something we should think about doing. This could create some confusion if we don't update language consistently across the app.

src/app/page/UserProfile/UserOverview.jsx
```diff
@@ -59,9 +59,11 @@ const UserStatus = ({ profile, setProfile, view }) => {
...
   };
 
-  if (loaded && !error) {
-    handleScriptLoad();
-  }
+  useEffect(() => {
+    if (loaded && !error) {
+      handleScriptLoad();
+    }
+  });
```
jwu910, line 74: useEffect should include at least an empty array as its second argument. If this effect is dependent on loaded or error changing, then those variables should be included in that dependency array

src/app/page/UserProfile/UserOverview.jsx
```diff
@@ -50,7 +50,7 @@ const UserStatus = ({ profile, setProfile, view }) => {
       type: ["(cities)"],
     };
     const addrElem = document.getElementById("address");
-    const googleRef = !!window && !!window.google ? window.google : undefined;
+    const googleRef = window.google;
```
jwu910, line 61: Without the ternary logic, I think this is an unnecessary variable assignment and it would be more clear to just explicitly use `window.google` instead of `googleRef`.

src/app/page/UserProfile/UserOverview.jsx
```diff
@@ -16,6 +16,14 @@ const useStyles = makeStyles((theme) => ({
...
     margin: "12px auto 8px auto",
   },
+  profileEditButton: {
+    backgroundColor: "transparent",
+    color: theme.color.darkBlue,
+    "&:hover": {
+      backgroundColor: "transparent",
+      color: theme.color.secondaryBlue,
+    },
+  },
```
jwu910, line 26: Were these styles already pre-existing in the MUI theme or any of the app level styles?

priyankpat: It was already pre-existing in the theme.

src/app/page/UserProfile/UserOverview.jsx
priyankpat: Got it. Addressed in 78c8ef5. I don't think the effect should be dependent on loaded or error as the fallback mechanism is to just allow freeform entry of address/location.

azmainamin:
@jwu910 Can you take a look if the requested changes were implemented or not? Then we can merge this in

src/app/page/UserProfile/UserOverview.jsx
```diff
@@ -15,10 +16,23 @@ const useStyles = makeStyles((theme) => ({
...
 }));
 
 const UserStatus = ({ profile, setProfile, view }) => {
   const classes = useStyles();
+  const [loaded, error] = useScript(
+    `https://maps.googleapis.com/maps/api/js?key=${process.env.REACT_APP_MAPS_ACCESS_TOKEN}&libraries=places`
```
jwu910, line 32: Do we have any documentation updates on where this new environment variable is coming from? If we're merging this in, this feature will not work if we're expecting another local environment variable.

src/app/page/UserProfile/UserOverview.jsx
```diff
@@ -29,6 +43,35 @@ const UserStatus = ({ profile, setProfile, view }) => {
...
+    }
+  };
+
+  const handleScriptLoad = () => {
+    const options = {
+      type: ["(cities)"],
+    };
+    const addrElem = document.getElementById("address");
+
+    if (window.google !== undefined) {
```
jwu910, line 62: If we're only checking for `!== undefined` this can we just check for anything truthy?

`if (window.google)` ?

priyankpat: Addressed in 07800cc

src/app/page/UserProfile/UserOverview.jsx
priyankpat: Addressed in 0d3f857. We should update the example .env file that is pinned to the slack channel to include `REACT_APP_MAPS_ACCESS_TOKEN`

src/app/page/UserProfile/UserOverview.jsx
jwu910: If `handleScriptLoad` is NOT dependent on `loaded` and `not error`, then there should be no reason to use these variables in an `if` block

src/app/page/UserProfile/UserOverview.jsx
jwu910: My question was maybe unclear.

Do we need a specific profileEditButton style object if our theme buttons already cover these styles.

https://material-ui.com/components/buttons/

Just from seeing the source here, it looks like you're trying to achieve the "text button" look in the MUI components.

priyankpat: Ah, yeah that styling was redundant. Addressed it in 15ebb3b.

src/app/page/UserProfile/UserOverview.jsx
priyankpat: Sorry, I misread the question. handleScriptLoad is dependent on `loaded` and `not error`.

README.md
```diff
@@ -26,6 +26,12 @@ npm ci
 
 Copy the file pinned to the side of #resilience-app channel to `.env.development`in the git root of this repo.
 
+- Example of .env.development content
+```
+REACT_APP_MB_ACCESS_TOKEN=
+REACT_APP_MAPS_ACCESS_TOKEN=
+```
+
```
jwu910, line 34: Can you include the algolia keys as well?

jwu910:
This is closed on AirTable, is this a will not fix?

/CC: @DAHuerta @utunga @mat10tng @qdozaq

qdozaq:
Yeh looks like this has been handled during some other location updates. Gonna close

PR #306, opened 2020-04-22
partially fixes #298

### LINKED ISSUE #298: Required fields on create mission page do not contain an "*"
When someone creates a mission, all fields (except the photo) are required.  However, not all fields have an "*".  For example, notice that the "Mission title" has a "*" but the "Add a recipient" does not ...

Screen Shot 2020-04-21 at 9 38 10 AM

This is true as well before someone enters text in the field, the placeholder text boxes do not consistently contain the required "*" characters ...

Screen Shot 2020-04-21 at 9 37 44 AM

ShanuCreations:
@laredotornado - Fixed the required fields as per #298 except 'Pickup' & 'Drop-off' as there is no required option to specify in Algolia Places API. I can fix this later if someone sheds any light here. I can't think of an alternative without impacting other areas.

### DIFF
--- a/src/app/page/MissionCreate/MissionForm.js
+++ b/src/app/page/MissionCreate/MissionForm.js
@@ -144,6 +144,7 @@ function MissionForm({ getFile, handleChange, onSubmit, values /*, assignHelper,
           onChange={handleChange}
           label="Assign a recipient"
           helperText="Name or Phone number"
+          required={true}
         />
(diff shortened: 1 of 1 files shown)


ShanuCreations:
@qdozaq - Thanks for your response ! It seems to be already doing the validation of Mission Form submission as per the function checkForEmptyInput under src\app\page\MissionCreate and throwing 
 a generic error on Submission.

Issue #260, opened 2020-04-17
As a temporary measure can we write some sort of CI integration to reset the db every so often??


utunga:
@mat10tng  has a great wee script that populates some fake data into our crazy shared development db set up (at firebase).

Obviously the right thing long term is that every developer has their own db and integrations or some sort of mock or something.. but as a temporary measure can we write a script that resets all the data in dev every so often? Like every 2 hours or something? 

I am not entirely sure where the code @mat10tng has been running actually is but maybe you can tell us or maybe we can find it. It was python once upon a time.

mat10tng:
ehm, it is in my branch still which is feat/story-116-missions-map-list-view
in the root directory
/scheme/data.py

rainrivas:
@mat10tng just so i'm understanding the flow here, you're running `data.py` to build your dataset, then you're running the `uploadMock.js` to push the data up to firefbase yes?

mat10tng:
Yeah you're right. Also, I wrote that in according to scheme.ts in the
model folder. But that scheme might change. I am just do manuall check for
now though

rainrivas:
@mat10tng @utunga  I'm getting pretty close to cracking this egg. Workingo n a fork to test out some github actions to execute the python script and then the nodejs script. I need to find out how we are meant to load secrets & environment variables so that the file so that we can node script can correctly update the DB.

rainrivas:
Here is a sample of the seed action running -- https://github.com/rainrivas/resilience-app/actions/runs/82457841

Currently clocked to run every minute but I will be changing it to run every 2 hours as @utunga suggested

utunga:
That looks super cool @rainrivas. I guess my only question is what to do from here to run this on our repo? Or rather our dev db? Is it already running?

rainrivas:
@utunga It's not currently running on our DB since I do not have the secret keys to update firebase. We should be able to get this running by creating the same workflow in this repo -- i can open a PR for this. I need to do a quick test on my fork for using Secrets from GH settings to pass into the workflow. If I get this working (will test tomorrow) I will connect with you guys to get the appropriate secret file added to the repo. I'll only need the secret name so I can configure `uploadMock.js` to use that to create the connection

utunga:
@rainrivas cool... creating a PR sounds good. Does github have a function for storing secrets in env variables in a secure way like the way that heroku and such do?

rainrivas:
[...]

PR #313, opened 2020-04-24
Addresses #260 -- seed db every 2 hours

rainrivas:
Addresses #260 
Creating a workflow (`seed.yml`) to fill our firebase with new data every 2 hours. 
**NOTE** The script will fail until `FIREBASE_SECRET` has been added to repo. I think @mat10tng  has already done this, but adding him on here to confirm


rainrivas:
Check error seems to be due to the node step on the CI not correctly downloading nodejs 

Use Node.js 10.x
32s
```
##[error]connect ETIMEDOUT 104.20.23.46:443
Run actions/setup-node@v1
  with:
    node-version: 10.x
    always-auth: false
##[error]connect ETIMEDOUT 104.20.23.46:443
```
If there is some way to rebuild it I'm sure it will pass

jwu910:
Re-ran the check

jwu910:
Accidentally closed the PR sorry. Finger slipped.

Was going to comment and ask, are the permissions set up to allow CI to write to the DB to re-seed with this job?

utunga:
lemme know if i can help?

azmainamin:
This looks good to me. I approved but I'd like someone with more knowledge of this to merge it in.

rainrivas:
@azmainamin so the seed.yml will install dependencies required to run the python script that @mat10tng wrote as mock data. 
It will also install the npm module for firebase and set the auth token using github secrets.  It then calls the node script Tuan wrote to publish our fake data to firebase.
@jwu910 the CI doesn't really write anything; it's the user in the secret that Tuan added as part of github secret in admin settings. The update happens via a POST call from what I'm gathering
```

async function upload(data, path) {
  return await admin
    .firestore()
    .doc(path.join("/"))
    .set(data)
    .then(() => console.log(`Document ${path.join("/")} uploaded.`))
    .catch(() => console.error(`Could not write document ${path.join("/")}.`));
}
```

PR #327, opened 2020-04-26
Setting up a single source to get the organization ID

qdozaq:


src/app/model/Mission.ts
```diff
@@ -43,9 +43,9 @@ const defaultMissionData: MissionInterface = {
   recipientId: "No Recipient Id", // reference?
 };
 
-const fsInProposed = {
+const fsInProposed = (orgId: string) => ({
```
qdozaq, line 46: Need to change these to functions otherwise no way to actually update the organization Id values.

I'm really not a big fan of how we're doing this here, but we will have to wait to restructure things as that could take a good chunk of time

PR #331, opened 2020-04-26
Feat/125 NOT A REAL PR! BUT A CRY FOR HELP

thirdeyeclub:
ha. besides the title, making this PR to see if anyone can help me with a part of the functionality I cannot figure out. Maybe someone can look at it and find a solution while I am finishing the rest of the functionality for the page. 

PROBELM: When i select mission type it creates a new input list that can have infinite items added to it, each element has its own state and validator, BUT i cannot get the value of the type of food box or the amount to display on the input after it is selected. This is a unique problem because the html element has its own state and functions created with it.

image

If anyone has an idea please comment or make a commit. 

lines that you will find the bug are as of commit 434 on this branch are 221 - 295

DO NOT WORRY ABOUT FINISHING THE REST OF THE FuNCTiONALITY, BUT IF YOU HAVE ANY OTHER COMMENTS I WILL GLADLY TAKE THEM I am only opening this PR for this one specific bug since I am stuck and this will make sure someone looks at it.


src/app/dashboard/Missions/CreateMission.jsx
```diff
@@ -0,0 +1,480 @@
...
+          </Select>
+        </FormControl>
+        {/* THIS PART SPAWNS BASED ON WHAT MISSION TYPE WAS SELECTED */}
+        {missionType === "food-box" ? (
+          <>
+            <Typography className={classes.inputHeader}>Select Food Box Options</Typography>
+            <FormControl className={classes.form}>
+              {/* DETERMINE FOOD BOX TYPES */}
+
+              {foodBoxInputCount.map((index) => {
```
qdozaq, line 227: Have you tried tried making this map to an actual component instead of an anonymous function with state?
I don't know for sure but I'm willing to bet react cant reconcile states correctly like this, not to mention it's not very performant.

jwu910:
Just pulled your code down, gonna poke around and see if i can find anything

jwu910:
Hey @thirdeyeclub This PR seems to have this form a little overengineered and I've been on n off reading the code for a few hours and it's been pretty hard to follow.

We have a lot of business logic in the jsx right now which is creating some really confusing function scopes which I think is making it a little harder to do what you're doing.

If you abstract some of the logic into more singular purposed functions, it might be a little easier to keep the multi-level state you're setting up with the box types + box counts.

I think breaking this down into smaller tasks might make figuring out the logic a little bit more.

LMK if you have questions and I can try to find some time to take a deeper look

mat10tng:
Ok, first, lets split it out to container pattern!
After that, maybe split out to different mission view depends on type. 

I think that would likely allow us to understand what is going on

PR #333, opened 2020-04-27
Fixes #302

### LINKED ISSUE #302: When creating a mission, back-end shouldn't force "Helper" field to be required
Currently when you create a mission, you are forced to fill out all fields (except photo).  Failing to fill out all fields causes validation to fail when you click "Create Mission".  However, "Helper" should not be required

Screen Shot 2020-04-21 at 7 47 43 PM

ShanuCreations:
I've applied the fix for Mission 'Helper' field back end validation as per #302. Could someone please review and let me know if any issues ?

### DIFF
--- a/src/app/page/MissionCreate/MissionCreate.jsx
+++ b/src/app/page/MissionCreate/MissionCreate.jsx
@@ -37,12 +37,12 @@ function MakeMission({ history }) {
       setLoading(true);
       //Lookup id of helper
       let volunteerId;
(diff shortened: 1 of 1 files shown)


ShanuCreations:
Thanks for the review @qdozaq ! Ah ok, I had prettier configured on my VS Code locally, hence the formatting was changed. I've sorted that out and merged the changes, could you please recheck ?

PR #337, opened 2020-04-27
Fixing "User.getIdByDisplayName" function

laredotornado:
This was previous returning the volunteer object instead of the ID, which is needed by the mission when creating it.


src/app/model/User.ts
```diff
@@ -72,10 +72,12 @@ class User extends BaseModel {
       id: missionId,
       ...mission,
     });
+    mission.id = missionId;
 
     //save mission in firestore
     try {
-      await collection.doc(missionId).set(sanitizedMission);
+      await collection.doc(missionId).set(mission);
+      //await collection.doc(missionId).set(sanitizedMission);
```
azmainamin, line 80: We were using this.load() to merge default data with the passed in data. If we do not use this, we don't create missions with default data.

laredotornado: @azmainamin , my bad!  I didn't mean for that to get checked in ... I was messing around and left that in by mistake.  Good catch.

src/app/model/User.test.ts
```diff
@@ -54,8 +126,8 @@ xdescribe("User", () => {
       mission.status = MissionStatus.assigned;
     });
 
-    it("unassigns volunteer if missionId exists", async () => {
-      const { mockData, mockDocFn, mockUpdate } = mockBaseRepo({
+    /*  it("unassigns volunteer if missionId exists", async () => {
```
azmainamin, line 129: Should these tests be commented out? Can we just remove them for now?

Issue #336, opened 2020-04-27
Mission workflow validation only enabled in detail view


jwu910:
**Background**
Phone number verification is required to accept and start missions when viewing a mission detailed view. Accepting or starting a mission from the mission list view should also be protected behind number verification.

**Acceptance Criteria**
All mission workflow actions should be protected against unverified users

![image](https://user-images.githubusercontent.com/12107969/80419205-8d438480-888d-11ea-9a93-7fee4d252056.png)

![image](https://user-images.githubusercontent.com/12107969/80419248-a1878180-888d-11ea-90fd-a46e143b6e6d.png)

PR #338, opened 2020-04-27
Bug/verify phone on mission list view

azmainamin:
Resolves Issue #336


azmainamin:
> 
> 
> LGTM with these changes, just quick question though. Should our exposed APIs like `User.startMission` be the place that checks if `currentUser.phoneNumber` is valid before the request is made?
> 
> Would it make more sense? Then anytime from any component `User.startMission` can return success or error

That's a great point! We probably should do some verification on our endpoint as well! I will create an issue in GH

Issue #339, opened 2020-04-28
Adding IE support inside of package.json for Resilience App


CSumm:
Tested app on IE for compatibility and it does not appear. Potential solution is to apply it to the Package.json development list of browsers to be supported.

jwu910:
For reference https://github.com/browserslist/browserslist

jwu910:
@CSumm Can test to make sure everything works after adding, if not, we may want to re-open this. Otherwise we should be good leaving it closed.

Thanks!

CSumm:
Tested and seems to persist, trying to implement react-app-polyfill along with the browserlist additions now

CSumm:
@jwu910 I managed to get it running after inserting the react polyfills into index.tsx and adding the broowserlist ie 11, but i get this error that appears 
![image](https://user-images.githubusercontent.com/31711543/80443526-6c0a8480-88dd-11ea-95e1-511812acaf3c.png)

*this browser does not support this* as the warning

jwu910:
Hoping to avoid the bloat of polyfills, we can revisit tomorrow and check with @utunga to see how much of IE we would be trying to support

PR #341, opened 2020-04-28
Add react-app polyfill for ie9 and ie11 into index.tsx, patching last fix for ie support

CSumm:
This PR fixes compatibility issues with IE11, allowing the application to run on the browser.
Relates to issue #339 and closed PR #340

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "mui-datatables": "^2.14.0",
     "path": "^0.12.7",
     "react": "^16.13.1",
+    "react-app-polyfill": "^1.0.6",
     "react-dom": "^16.13.1",
     "react-facebook-login": "^4.1.1",
     "react-firebaseui": "^4.1.0",
--- a/package.json
+++ b/package.json
@@ -84,6 +84,7 @@
       "not op_mini all"
     ],
     "development": [
+      "ie 11",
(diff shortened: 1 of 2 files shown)


jwu910:
Can we use browserslist to attempt to build for latest IE build and not use polyfills? CRA should be reading the brwoserslist and adjust the builds accordingly as much as possible. 

If we lose some functionality on latest IE, we may just want to have that discussion about not supporting it should some of the features be not worth the time

CSumm:
I reread the create-react-app documentation once more and under the browserlist example, there's an info section stating that we would need polyfill support as well as adding to browserlist: https://create-react-app.dev/docs/supported-browsers-features/#configuring-supported-browsers. This can be found near the end of the page.

This means that to support IE, polyfills are necessary it seems.

jwu910:
I haven't checked any of this on windows yet @CSumm.
But yeah reading through the docs you've shared it seems so. I would look to checking in with the team to have that discussion if ie11 is something we want to support or just direct users to installing one of the modern browsers.

Looking at some of the polyfill required features, we are using a couple of them.

CSumm:
Sounds like a good idea and wait for more opinions on this. For now, we could put this PR on the back burner.

PR #345, opened 2020-04-28
Fix wet code in typography.jsx, use MuiTheme object instead of individual typography variant wrappers

CSumm:
Commented out the wet code so it can be kept in if necessary, added code that utilizes CreateMuiTheme which exports the theme object that can be used for typography variants.

### DIFF
--- a/src/app/component/Typography.jsx
+++ b/src/app/component/Typography.jsx
@@ -1,40 +1,60 @@
-import { Typography } from "@material-ui/core";
-import React from "react";
+//import { MuiTypography, variantMapping } from "@material-ui/core";
+//import React from "react";
+import { createMuiTheme } from "@material-ui/core/styles";
 
-const H1 = ({ children, ...rest }) => (
-  <Typography variant="h1" {...rest}>
-    {children}{" "}
-  </Typography>
-);
-const H2 = ({ children, ...rest }) => (
-  <Typography variant="h2" {...rest}>
-    {children}
-  </Typography>
-);
-const H3 = ({ children, ...rest }) => (
-  <Typography variant="h3" {...rest}>
-    {children}{" "}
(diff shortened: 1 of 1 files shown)


CSumm:
I noticed that home.js relies on Typography.jsx and resulted to a failed test due to changing to dry code in Typography. Should I go ahead and make necessary changes to home.js to reflect changes in typography variant mapping along with this PR?

jwu910:
Yeah you should have implementation changes as well. Don’t want this breaking the app for everyone else if we merge half the work.

I haven’t looked at the changes yet, just seeing your messages.

We’ll get it reviewed after the implementation details are done too

src/app/component/Typography.jsx
```diff
@@ -1,40 +1,60 @@
...
+// const Body2 = ({ children, ...rest }) => (
+//   <Typography variant="body2" {...rest}>
+//     {children}{" "}
+//   </Typography>
+// );
 
-export { H1, H2, H3, H4, H5, Body1, Body2 };
+// export { H1, H2, H3, H4, H5, Body1, Body2 };
+
+export const theme = createMuiTheme({
```
jwu910, line 43: Just looking at the changes you're suggesting here, by the looks of the file, we're expecting to export Typography components. The change you've made is exporting a theme object.

This definitely will break implementation of this file. I'm 95% sure the variantMapping object you're adding in to the createMuiTheme method should be in the app level theme object that we're probably setting up somewhere else in the application.

CSumm: I realize that my implementation was incorrect as well. I attempted to try and keep it somewhat similar to the original only to realize I can't return an object inside of a const method as I expected. I think what I will do is close the PR and try to rethink the strategy. What do you think?

PR #347, opened 2020-04-29
Fix get all available missions

azmainamin:
Available == status of "tentative"


qdozaq:
Oh. looks like that test needs to be updated since that method was removed

PR #349, opened 2020-04-29
Fixes #344

### LINKED ISSUE #344: [BUG] Success message is missing when user updates profile on user profile page
1. Go to app using localhost or the Heroku development app
2. Sign up or sign into app whichever way
3. Open up navigation sidebar and go click on user profile.
4. Make a change to display name or add an image
5. Observe result once change is made

Expected result: Success message appears when editable item inside of user profile is changed.

Actual result: Success message is missing when an editable item is changed inside of user profile.

Relates to Airtable card 75: https://airtable.com/tblRjd2McKgRW9MKj/viwlRRNI0PZIxrGkc/recqdAylRmM3FwpAJ?blocks=hide

ShanuCreations:
Success message added when user updates profile as per #344 . Let me know if any issues.

### DIFF
--- a/src/app/page/UserProfile/UserProfile.js
+++ b/src/app/page/UserProfile/UserProfile.js
@@ -8,7 +8,7 @@ import { useFirebase, useFirestore } from "react-redux-firebase";
 import { withRouter } from "react-router-dom";
 
 import { Button } from "../../component";
-import ErrorSnackbar from "../../component/Snackbars/ErrorSnackbar";
+import { ErrorSnackbar, SuccessSnackbar } from "../../component/Snackbars";
 import { Card, Page } from "../../layout";
 import addressLookUp from "../../utils/addressLookUp";
 import LinkGoogleAccount from "./LinkGoogleAccount";
@@ -85,6 +85,7 @@ const UserProfile = ({ history }) => {
               addressMapPoint: { latitude: data.lat, longitude: data.long },
             })
           );
+      setSuccessSnackbarOpen(true);
     } catch (error) {
       console.log("ERROR WHEN GETTiNG LOCATION", error);
     }
@@ -131,6 +132,9 @@ const UserProfile = ({ history }) => {
     return data.providerId === "phone";
   });
 
+  /*=== Snack Bar - Success Message ===*/
+  const [successSnackbarOpen, setSuccessSnackbarOpen] = useState(false);
+
   // === Error Messages ==== //
   const [notSupportedError, setNotSupportedError] = useState(false);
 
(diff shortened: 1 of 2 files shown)


src/app/page/UserProfile/UserProfile.js
```diff
@@ -85,6 +85,7 @@ const UserProfile = ({ history }) => {
               addressMapPoint: { latitude: data.lat, longitude: data.long },
             })
           );
+      setSuccessSnackbarOpen(true);
     } catch (error) {
       console.log("ERROR WHEN GETTiNG LOCATION", error);
```
azmainamin, line 90: Can we add the error snack bar here where we cannot get location and remove this console log?

azmainamin:
@ShanuCreations  Just had 1 comment. Also, can you fix the npm run format:check warnings thats causing the build to fail?

ShanuCreations:
@azmainamin - Thanks for the review ! I already ran npm run format prior to creating PR, despite that it fails. I guess this is a known issue with our build as even for my previous PR @qdozaq seems to have ran npm ci manually, which fixed the issue.

src/app/page/UserProfile/UserProfile.js
ShanuCreations: @azmainamin - I can certainly do that, but the issue is there is already <ErrorSnackbar> setup on that page for an existing user account with different error message, and this logic needs to be changed a bit to avoid duplication. Perhaps, it requires a separate issue on it's own ?

azmainamin:
@ShanuCreations I approved it but can you remove unused "useSelector" from Appbar.js? That would be awesome!

ShanuCreations:
@azmainamin - Thank you ! Sure, done and pushed the changes to the same branch. Could you please check ?

jwu910:
@ShanuCreations Just a heads up, the `ErrorSnackbar` and `SuccesssSnackbar` is being deprecated in favor of the `Snackbar` component. There should be deprecation warnings in the console when you are viewing these. It should be have the same, the api is just set up a little more generic to accept both success and error messages.

If you want, you can re-open the issue and submit another PR to update that. It's not a blocker, just part of our cleanup efforts to reduce wet code.

Thanks!

azmainamin:
@jwu910 Great catch! Maybe we should delete Error and Success snackbars so people don't get confused?

jwu910:
I added deprecation warnings, but didn't want to go in and re-do the already implemented Snackbars.

Some components had some complex logic and callbacks rendering each snackbar, and i didnt have time to figure out what was going on.

PR #350, opened 2020-04-29
Update README with instruction for setting .env locally.

marekdano:
@utunga, please review


jwu910:
What is the `.env` folder?

We should be okay with only a file `.env.development`

To support the documentation as well too. we can add in a `.env.sample` file with a comment to rename the file. And this file can keep the left hand side keys of all of the env variables we need to run the app.

PR #400, opened 2020-05-02
Fix markdown links on README

utunga:
Make it so they render OK in the github conext where they are by far most likely to be seen.


utunga:
To get a preview visit here (and scroll down to File structure section) 
https://github.com/factn/resilience-app/tree/docs/fix-readme 

Compare and contrast with master

utunga:
OK. So in the end I just used the full url. But according to what I read on stackoverflow this might even be preferable. " After all, if someone has downloaded your repo and is viewing the source locally, the relative URL to the Repo root will be different than when viewing the file on GitHub. In that case, you probably want to point them to GitHub anyway."

Besides which - and most importantly - it actually works.

jwu910:
That should be fine, doc links should be pointing to versions controlled by the master branch :+1:

README.md
```diff
@@ -166,32 +166,32 @@ git push heroku master
...
+| File                                                                                                             |                                                                Description                                                                 | Edit? |
+| ---------------------------------------------------------------------------------------------------------------- | :----------------------------------------------------------------------------------------------------------------------------------------: | ----: |
+| [config/](https://github.com/factn/resilience-app/tree/master/src/config/)                                       |    This is used to build out the webpack. In most cases, this won't need to be edited unless we decide to change our build in some way.    |     N |
+| [node_modules/](https://github.com/factn/resilience-app/tree/master/node_modules/)                               |               The folder for all of the modules pulled in by Node. Good folder to see what FontAwesome icons are available.                |     N |
```
jwu910, line 172: This never gets committed. Link will always be broken

README.md
```diff
@@ -166,32 +166,32 @@ git push heroku master
...
+| [src/js/pages/](https://github.com/factn/resilience-app/tree/master/src/js/pages/)                               |                    This folder is for all of the JSX components that wrap full pages and are used as Routes in `App.jsx`                    |     Y |
+| [src/js/resources/](https://github.com/factn/resilience-app/tree/master/src/js/resources/)                       |                            This is general JavaScript that helps the rest of this code do what it needs to do.                             |     Y |
+| [src/scss/](https://github.com/factn/resilience-app/tree/master/)                                                |                                                             All Sass partials                                                              |     Y |
+| [src/scss/global/](https://github.com/factn/resilience-app/tree/master/src/scss/)                                |                                Initial tag settings, set globally before use. Similar to standardized CSS.                                 |     Y |
```
jwu910, line 184: Doesn't exist. project went a different route with the styling

README.md
```diff
@@ -166,32 +166,32 @@ git push heroku master
...
+| [src/scss/layout/](https://github.com/factn/resilience-app/tree/master/src/scss/layout/)                         |         This puts the general block elements on the page where they belong and dictate the structure of pages in a general sense.          |     Y |
+| [src/scss/object/](https://github.com/factn/resilience-app/tree/master/src/scss/object/)                         |                                This is for objects that are rendered on multiple pages in multiple contexts                                |     Y |
+| [src/scss/pages/](https://github.com/factn/resilience-app/tree/master/src/scss/pages/)                           |                                          This is for the styles specific to one page on the site                                           |     Y |
+| [src/scss/presets/](https://github.com/factn/resilience-app/tree/master/src/scss/presets/)                       |                  These have variables, mixins, and other presets for how the CSS is written so that it's faster to build.                  |     Y |
```
jwu910, line 188: Can probably delete these too

README.md
utunga: FWIW I found it helpful to have  a bit of a guide to where things should be when trying to get familiar with things.. 

That said I guess if it's wrong it does more harm than good.

utunga: I could live with just deleting this whole section I guess...

PR #401, opened 2020-05-02
fix data.py where i forgot comma after choice create wrong data

mat10tng:


utunga:
Oh I see! its not that you forgot a comma you added a comma and forgot to remove it ! Turning it into an array! tx

PR #403, opened 2020-05-02
Feat/147 download csv

utunga:
Adds two download links to the dashboard home 

Does not match figma but the figma is MVP1 I think not MVP0.


src/app/dashboard/Home/OverviewItem.jsx
```diff
@@ -26,7 +25,7 @@ const HomeItem = ({ icon, link, linkLabel, nbr, textColor, title }) => {
             <H3 style={{ color: textColor }}>{title}</H3>
```
mat10tng, line 25: what textColor is this ? why dont we use theme from useStyles here ?

utunga: sounds good to me - fwiw this is not code i wrote though.. was here before the PR.. its a property set from above - so planning is one color, in progress another etc.

src/app/model/schema.ts
```diff
@@ -159,11 +159,13 @@ export interface FoodBoxDetails {
 export interface MissionInterface {
   id: string;
   type: MissionType;
-
+  createdDate: string; // TODO should be a date?
```
qdozaq, line 162: been a bit since I talked with Tuan about it, but I think we should store dates as a unix timestamp integer. `(new Date()).getTime()`

Storing date objects always turns into a nightmare.

PR #405, opened 2020-05-02
Adding some documentation around test data

amylo:


amylo:
> Lgtm; But should we be explicit about python version?
> 
> Also does anyone else run this besides @mat10tng ?

I tested that it worked on both Python v2 and v3. So I didn't think we'd need to explicitly state the version since there were no issues.

jwu910:
Sounds good to me then

utunga:
FWIW I know this has been merged already but FYI I am pretty sure we added a thing that was supposed to run this every hour or so.. as a devops thing.

PR #406, opened 2020-05-03
Feat/144 [WIP]

qdozaq:
# WORK IN PROGRESS

### DIFF
--- a/src/app/component/AddressInput/AddressInput.jsx
+++ b/src/app/component/AddressInput/AddressInput.jsx
@@ -5,7 +5,7 @@ import React from "react";
 const { ALGOLIA_API_KEY, ALGOLIA_APP_ID } = process.env;
 
 const AddressInput = (props) => {
-  const { placeholder, setStage, stage } = props;
+  const { error, onClear, placeholder, setStage, stage } = props;
   const handleLocation = (query) => {
     if (query.suggestion) {
       const { countryCode, county, latlng, value } = query.suggestion;
@@ -22,19 +22,22 @@ const AddressInput = (props) => {
     }
   };
   return (
-    <AlgoliaPlaces
-      placeholder={placeholder || "Search address"}
-      name={stage}
-      options={{
-        appId: ALGOLIA_APP_ID,
-        apiKey: ALGOLIA_API_KEY,
-        language: "en",
-        countries: ["us"], // we have to support more countries in the future
-        // Other options from https://community.algolia.com/places/documentation.html#options
-      }}
-      onChange={(query) => handleLocation(query)}
-      onLimit={({ message }) => message && console.log(message)}
-    />
+    <div style={error ? { border: "solid red" } : {}}>
+      <AlgoliaPlaces
+        placeholder={placeholder || "Search address"}
+        name={stage}
+        options={{
+          appId: ALGOLIA_APP_ID,
+          apiKey: ALGOLIA_API_KEY,
+          language: "en",
+          countries: ["us"], // we have to support more countries in the future
+          // Other options from https://community.algolia.com/places/documentation.html#options
+        }}
+        onChange={(query) => handleLocation(query)}
+        onLimit={({ message }) => message && console.log(message)}
+        onClear={onClear}
+      />
+    </div>
   );
 };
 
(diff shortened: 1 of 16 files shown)


src/app/page/Request/foodbox/ConfirmStep.jsx
```diff
@@ -53,16 +60,48 @@ function ConfirmStep({ dispatch, state }) {
...
-    if (isDonationRequest) {
-      //TODO set the funding status to not funded
-      //Send mission to firestore
+  // TODO we should try to create this mission before paying and if payment fails we delete it or something
+  async function confirmRequest() {
+    const { cart } = state;
+    let mission = {
+      type: MissionType.foodbox,
+      // what does this need to be??
+      // status: ???
```
qdozaq, line 69: What does the status need to be for a new mission?

src/app/model/User.ts
```diff
@@ -72,34 +69,6 @@ class User extends BaseModel {
     return this.getCollection("users").where("displayName", ">=", value).limit(limit).get();
   }
 
-  /**
-   * Given a mission object creates a new mission in firestore
-   * returns the new mission id
-   * @param {object} mission
-   * @return {string}
-   */
-  async createMission(mission: MissionInterface): Promise<string> {
```
qdozaq, line 81: Moved this method to `Mission.ts` as it makes more sense there is actually associated with the correct default data

qdozaq:
Running into an issue where users that have signed in via a social account still need to input their phone number. Should I just handle in another PR?

src/app/component/AddressInput/AddressInput.jsx
```diff
@@ -43,6 +59,8 @@ AddressInput.propTypes = {
   placeholder: PropTypes.string,
   stage: PropTypes.any, // pickUp or dropOff
   setStage: PropTypes.func,
+  error: PropTypes.bool,
+  onClear: PropTypes.func,
```
qdozaq, line 63: reminder for myself add value here

src/app/component/AddressInput/AddressInput.jsx
```diff
@@ -1,11 +1,16 @@
...
 import PropTypes from "prop-types";
-import React from "react";
+import React, { useRef, useEffect } from "react";
 
 const { ALGOLIA_API_KEY, ALGOLIA_APP_ID } = process.env;
 
 const AddressInput = (props) => {
-  const { placeholder, setStage, stage } = props;
+  const { error, onClear, placeholder, setStage, stage, value } = props;
+  const placesRef = useRef();
```
qdozaq, line 9: Adding this ref so we can set a default value in the algolia input.
Needed for when a person goes back in their checkout experience to change details

src/app/page/Request/foodbox/ConfirmStep.jsx
utunga: new mission should have status 'Unassigned'  if its paid for by recipient it pops straight into 'funded' therefore 'tentative' (but I would do that *after* the payment is confirmed.)

utunga:
> Running into an issue where users that have signed in via a social account still need to input their phone number. Should I just handle in another PR?

FWIW there was a whole story about that and  PR added (and merged) that if someone tries to volunteer for a mission and they haven't got a phone number then they get asked to add their phone number first. 

The reasoning for this is we don't want to stand in the way of people getting excited and 'joining up' but otoh we do need a phone number before they can actually be assigned a mission.

I don't wanna hit them with a 'confirm your phone number' when they sign in..

src/app/page/Request/foodbox/DeliveryStep.jsx
```diff
@@ -1,25 +1,40 @@
...
 import NavigationButtons from "./NavigationButtons";
 import { User } from "../../../model";
 import { useForm } from "../../../hooks";
 
+const RECAPTCHA_ID = "recaptcha_id";
+
 function DeliveryStep({ dispatch, state }) {
   const classes = useStyles();
   const auth = useSelector((state) => state.firebase.auth);
+  console.log("auth", auth);
```
mat10tng, line 25: log.

PR #410, opened 2020-05-03
Fix create mission page

marekdano:
The `create a mission` page was broken. I had to import `firebase/storage` into `MissionCreate` component.

I found that there is no tests for testing this component. My suggestion is to create unit tests or e2e tests to test `create a mission` page


jwu910:
Looking through the history of the file. I think I actually broke this on April 19th. Have we not created a mission since then?

Can we just import this at App.jsx?

marekdano:
Just moved it to App. 
I can confirm that the page is broken on `heroku` too. It seems that nobody uses it or hasn't been sure what happened there when they wanted to create a mission.

PR #411, opened 2020-05-03
Formatting new header for Mission Card

laredotornado:
Going in small steps here.  Basically, trying to get the title section right from individual cards, as could be seen here

volunteer_image


src/app/component/MissionCard.jsx
```diff
@@ -53,7 +60,7 @@ const MissionCardContent = ({ classes, contentItems }) => (
  * @component
  */
 const MissionCard = withStyles(styles)(({ children, classes, mission, ...rest }) => {
-  const title = mission.title;
+  const title = mission.title ? mission.title : "No title supplied.";
```
jwu910, line 63: Pattern related: This should be syntatically similar to the pattern already established below https://github.com/factn/resilience-app/pull/411/files#diff-5261dfc248a50fac298f589e95348d05R65
```
mission?.title || "No title supplied."
```

Bigger question: Should this be a required field? Can a mission be created without a title?

React suggestion: We can leverage [defaultProps](https://reactjs.org/docs/typechecking-with-proptypes.html#default-prop-values) for this component. We already have the proptypes established below, so something like
```
MissionCard.defaultProps = {
  mission: {
    title: "No title supplied",
    status: "",
    pickUpLocation: {
      address: "No Data"
    }
  }
};
```

Default props will be the assigned values to props unless one is given during component creation

jwu910: This can clean up some of the logic in this component

src/app/component/MissionCard.jsx
```diff
@@ -76,11 +83,19 @@ const MissionCard = withStyles(styles)(({ children, classes, mission, ...rest })
...
     <Card className={classes.root} {...rest}>
-      <CardHeader
-        title={title}
-        titleTypographyProps={{ variant: "h2", component: "span" }}
-        className={classes.cardHeader}
-      />
+      <CardContent className={classes.cardContent}>
+        <Grid container spacing="1" alignItems="flex-start" justify="flex-end" direction="row">
+          <Grid item>
+            <img height="20" src={require("../../img/apple.svg")} alt="" />
```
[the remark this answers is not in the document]
laredotornado: Done!  Working on your other comment ...

src/app/component/MissionCard.jsx
laredotornado: Hi @jwu910 , I have added your suggestions.  Although it seems fairly obvious to me that "title" should be a required field, there seems to be tons of data in firebase without a title, making me wonder if it's possible to set a not-empty constraint at that level.  @mat10tng manages the models and could better speak to this requirement.

jwu910: Yeah would be good to discuss. We can always program defensively at least on the build time with typescript and/or utilizing propTypes for runtime errors to catch this stuff.

src/app/component/MissionCard.jsx
```diff
@@ -89,6 +105,16 @@ const MissionCard = withStyles(styles)(({ children, classes, mission, ...rest })
...
 
+MissionCard.defaultProps = {
+  mission: {
+    title: "No title supplied",
+    status: "",
+    pickUpLocation: {
+      address: "No Data",
+    },
+  },
+};
```
jwu910, line 116: This shape doesn't match what's being destructured from props.

The code sample i pasted was just a sample and I didn't write out the entire props object.

laredotornado: I'm sorry @jwu910 , I'm not clear on what you want me to do.  "address" is a field of the Location object, per the [schma](https://github.com/factn/resilience-app/blob/master/src/app/model/schema.ts#L5) .  Do you want me to populate default values for all fields that are part of a mission?

jwu910: Default props aren't necessary, but its a feature that allows us to pre-define values for props going into the components.

Looking again, i think I misread something. The shape looks right, but if we're going to set default values for `title` and `address`, would we need default values for the rest of the properties in the missions prop?

laredotornado: @jwu910 , I think that is a question that is above my paygrade!  :)  I would say, as an outsider looking in, I don't think defaultProps are needed at all, in as far as all these fields should be populated with something otherwise it wouldn't make sense to a volunteer (or a recipient).  @azmainamin may have some thoughts as well.

jwu910: You seem to be doing just fine lol.

What I was trying to get at was that PropTypes can be used at the component level to determine the data going in. If that data is coming from the backend, and its missing, thats a whole different problem. If its required on the backend, we can expect it on the front, the propTypes are here to help let the component know when props should be required.

Maybe going down a rabbit hole now:

Say I write a component:
```
const personDetail = (props) => {
  return (
    <div>
      {props.people.map(person => {
        return <div style={{color: props.color}}>
          {person.name}
          {person.age}
        </div>
      })}
    </div>
}
```

And you as another developer will be using `personDetail` in your component, you may write out:

```
<personDetail color='blue' />

// or

<personDetail people={[{name: 'Dave'}]} color='red' />
```

And because the component does not check to see if any values are missing when it's invoked, you're gonna throw a reference error(?) for people not a property of props, or age is not a property of person.

PropTypes and defaultProps(if we want them) help avoid some of these issues by giving the component a meaningful error message something like "`people is a required prop of personDetail, but none was provided`"

I think you're right on us not needing them per se, would be a whole different issue if they were missing.

azmainamin: What Josh said. I am not the biggest fan of default props. Its very easy to hide a bug (BE not sending the correct data) with a default Prop. I would use them if not having the data will break some logic: `const myVar = obj.prop` will break if obj is undefined. 

In this case, I have no problem not using default props.

src/app/component/MissionCard.jsx
utunga: My 2c on 'title' for missions is that they never made sense in the first place. 

A designer added it to a sketch at some point and it got copied everywhere but imho nobody stopped to ask who the heck is going to want to stop long enough to give a mission a 'title'? 

I worked with Karen to revise the designs and we tried to 'squash' the idea of a title on missions at that point but I don't know if maybe the idea lives on in the code. Like a zombie wandering the moors. 

Far as I am concerned the <MissionType> plus <Summary of mission details> can be used wherever a mission 'title' might be needed. 

For instance if its a food box mission the UI shows "FoodBox' and "3 Vegetable Fruit Medly" or just Foodbox: 3 depending on exact context. 

If its a 'generic errand' we may want to show just the first 40 characters of the full 'notes' field as a summary (that can be clicked on for more detail). But since generic errand is not in MVP0 anymore, we can cross that bridge when we come to it.

This doesn't relate to mission props but just title. 

I know that the code has diverged from the designs but I suggest we refer back to the designs and designers in terms of this 'title' thing. 

Also, related, the idea of having an 'image' for every mission. Nice in theory but doesn't make sense till after they are delivered. So.. think that's gone in the latest designs also.

src/app/component/MissionCard.jsx
laredotornado: @jwu910 @azmainamin , ok I have removed the default props.  I have also added in some more card content, in the interest of moving these features along.  Let me know what you think.

src/app/component/MissionCard.jsx
```diff
@@ -53,9 +66,10 @@ const MissionCardContent = ({ classes, contentItems }) => (
  * @component
  */
 const MissionCard = withStyles(styles)(({ children, classes, mission, ...rest }) => {
-  const title = mission.title;
+  const title = mission?.title || "No title supplied.";
```
utunga, line 69: Looks like the title came out of the CardHeader / Card but do we still want it in the View? aka this title property?

src/app/component/MissionCard.jsx
```diff
@@ -76,13 +90,72 @@ const MissionCard = withStyles(styles)(({ children, classes, mission, ...rest })
 
   return (
     <Card className={classes.root} {...rest}>
-      <CardHeader
-        title={title}
-        titleTypographyProps={{ variant: "h2", component: "span" }}
```
utunga, line 81: Not having the title here is good.. i think it matches the designs, yeah?

src/app/component/MissionCard.jsx
utunga: FWIW MissionType schema 
https://github.com/factn/resilience-app/blob/5930fb6164747cd58e4d40d078300149e92be8be/src/app/model/schema.ts#L118

But in fact we only supporting one mission type for MVP.0 - MissionType.foodbox

So i guess a magic string 'foodbox' is OK. But better to use the enum..

src/app/component/MissionCard.jsx
```diff
@@ -76,13 +90,72 @@ const MissionCard = withStyles(styles)(({ children, classes, mission, ...rest })
...
-        className={classes.cardHeader}
-      />
       <CardContent className={classes.cardContent}>
-        <MissionCardContent contentItems={contentItems} classes={classes} />
+        <Grid container spacing="1" alignItems="flex-start" justify="flex-end" direction="row">
+          <Grid item>
+            <img height="20" src={appleIcon} alt="" />
+          </Grid>
+          <Grid item style={{ flex: 1 }}>
+            {title}
```
utunga, line 99: image

Looking at the design it'd be great if we could get write a method that goes from FoodBoxDetails to something like "2 x Fruit & Veggies Medly" instead of just "No title supplied." here. 

From this data structure to a string. Tricky because we don't know what kind of details but we can put a method of "Mission" that generates a title, maybe? And then a switch on MissionType that then assumes if MissionType.FoodBox that we will have a FoodBoxDetails schema..

https://github.com/factn/resilience-app/blob/57030727a133e69ba4eca24520211c426268447c/src/app/model/schema.ts#L151

--

The complication is we may have more than one type of box. In practice, for now, it will always be the same one but to handle it in a general way maybe we can just do a "..." type thing.

So generate a title like "2 x Fruit & Vege Medly, 1 x Care Pack" but then cut it off so you get "2 x Fruit & Vege Med..." as the title? 

--

Also I think it'd be great to get the background to be grayed out or similar - we want people to click on the title  I think -  it will make it much more usable ?

src/app/component/MissionCard.jsx
```diff
@@ -76,13 +90,72 @@ const MissionCard = withStyles(styles)(({ children, classes, mission, ...rest })
...
+          </Grid>
+          <Grid item>
+            <InfoIcon />
```
utunga, line 102: I assume from the design that this is supposed to take you to the same place as "View Details" does in this button currently? 

Cause in the latest design there is no "View Details" button - making space for "unassign me" and other things.

I think the idea is that the user can press anywhere in the title, or ideally almost anywhere in the bulk of the card.. to get to the details?

src/app/component/MissionCard.jsx
```diff
@@ -76,13 +90,72 @@ const MissionCard = withStyles(styles)(({ children, classes, mission, ...rest })
...
+          <Grid item className={classes.halfRow}>
+            <Grid container direction="column">
+              <Grid item>
+                <Grid container direction="row">
+                  <Grid item>
+                    <PublishIcon />
+                  </Grid>
+                  <Grid item>
+                    <Typography variant="h5" noWrap>
+                      PICK UP
```
utunga, line 117: It'd be great if we could make these be green / purples per the design - just so they fade into the background a bit more I gueess.

src/app/component/MissionCard.jsx
```diff
@@ -76,13 +90,72 @@ const MissionCard = withStyles(styles)(({ children, classes, mission, ...rest })
...
+                    </Typography>
+                  </Grid>
+                </Grid>
+              </Grid>
+              <Grid item>{mission.pickUpLocation.label}</Grid>
+              <Grid item>
+                <a href={`https://www.google.com/maps/dir/"${location}"`}>{location}</a>
```
utunga, line 124: Is it possible to 'truncate' the location text so that it ends in "..." if its too long? (per the design)

karenkun: @laredotornado  [Material UI Typography](https://material-ui.com/api/typography/): `noWrap` true

laredotornado: @utunga @karenkun, Changes are in to create the ellipsis when text would normally wrap.

PR #412, opened 2020-05-03
Feat/model consolidation

mat10tng:


src/app/model/User.ts
```diff
@@ -45,246 +43,72 @@ class User extends BaseModel {
...
-    }
-    const missions = missionsDocumentSnapshot.map((doc) => doc.data());
-
-    return missions;
+  getAllCompletedMissions(userUid: string) {
+    return this.getCollection("organizations")
+      .doc(Organization.uid)
+      .collection("missions")
+      .where("volunteerUid", "==", userUid)
+      .where("status", "in", [MissionStatus.succeeded, MissionStatus.failed])
```
azmainamin, line 104: Did we remove "delivered" from the status? (I think we should).

src/app/model/User.old-test.ts
```diff
@@ -0,0 +1,242 @@
+// import ReduxFirebase from "react-redux-firebase";
```
qdozaq, line 1: Just deprecating this test. I don't think anything it tests even exists any more

src/app/page/Home/Home.js
```diff
@@ -372,14 +374,15 @@ const SignInHeaderComponent = ({ history }) => {
  *
  * @component
  */
-const HomePage = ({ auth, history }) => {
+const HomePage = ({ auth, history, profile }) => {
   const classes = useStyles();
-  const isEmpty = useSelector((state) => state.firebase.auth.isEmpty);
-  const isLoaded = useSelector((state) => state.firebase.auth.isLoaded);
+  if (isLoaded(auth) && !isEmpty(auth) && isLoaded(profile)) {
+    // User.createProfile(auth.uid, auth);
```
qdozaq, line 380: idk what the goal was here but I had to comment this out cause it was completely breaking the site if you weren't logged in

src/app/model/User.ts
qdozaq: looks like it's still in the schema

qdozaq:
This is a monstrosity......
User.test.ts was just completely broken and trying to call methods that didn't exist. Just got rid of it for now.

I fixed the weird UI issues with address field in the user profile.
Fixed some things that changed with the payment flow.

I was looking back at the user docs for react-redux-firebase and I think we're saving our users wrong.

See here: http://react-redux-firebase.com/docs/recipes/profile.html#update-profile
I think we need to be using this instead of just copying whatever is in `auth` into a user object in the db. Or is it the same thing?

src/app/model/User.ts
azmainamin: Then, "Missions Completed" should include missions that have been delivered, cause technically, they are finished and successful.

src/app/component/AddressInput/AddressInput.jsx
```diff
@@ -31,34 +26,35 @@ const AddressInput = (props) => {
...
-          countries: ["us"], // we have to support more countries in the future
-          // Other options from https://community.algolia.com/places/documentation.html#options
-        }}
-        onChange={(query) => handleLocation(query)}
-        onLimit={({ message }) => message && console.log(message)}
-        onClear={onClear}
-        placesRef={setRef}
-      />
+      {disabled ? (
+        <input className="ap-input" disabled="true" value={value} />
```
qdozaq, line 34: When the component doesn't have a disabled so you just steal the styles and conditionally render a disabled input. Works flawlessly lolol

src/app/model/User.old-test.ts
```diff
@@ -0,0 +1,242 @@
...
+//         throwUpdateError: true,
+//       });
+
+//       await expect(users.unvolunteerMission(missionUid)).rejects.toThrow(Error);
+
+//       expect(mockDocFn).toBeCalledWith(missionUid);
+//       expect(mockData).toBeCalledTimes(1);
+//     });*/
+//   });
+// });
```
jwu910, line 242: 😿

src/app/page/Request/foodbox/ConfirmStep.jsx
```diff
@@ -92,7 +92,8 @@ function ConfirmStep({ dispatch, state }) {
     }
 
     try {
-      await Mission.create(user.uid, mission);
+      const createdMission = await Mission.create(mission);
+      console.log(createdMission);
```
jwu910: Got a console log here

src/app/model/Mission.ts
```diff
@@ -208,63 +210,145 @@ class Mission extends BaseModel {
...
       });
   }
 
+  /**
+   * create a new mission
+   * returns the new mission id
+   * @param {object} mission
+   * @return {string}
+   */
+  create(mission: MissionInterface) {
```
qdozaq, line 246: I had to change this method because if we just use `add` it doesn't set the actual `uid` property on the mission object.
Turns out you can get an auto generated id by just generating an empty doc reference and then setting it. (see below)

src/app/model/Mission.ts
```diff
@@ -208,63 +210,145 @@ class Mission extends BaseModel {
...
+      .collection("missions")
+      .doc();
+
+    const newMission = this.load({
+      ...mission,
+      uid: newRef.id,
+      createdDate: Date.now().toString(),
+    });
+
+    return newRef.set(newMission).then(() => newMission);
```
qdozaq, line 259: returning the newly created mission when its done we so we have a reference for after

utunga:
> I was looking back at the user docs for react-redux-firebase and I think we're saving our users wrong.
> 
> See here: http://react-redux-firebase.com/docs/recipes/profile.html#update-profile
> I think we need to be using this instead of just copying whatever is in `auth` into a user object in the db. Or is it the same thing?

I'm not sure if this is the same thing but we wanted to do a thing that Firebase doesn't seem to do out of the box namely make the User model (in Firestore)  separate from the Auth in Firebase/auth. 

Specifically, Tuan did a ton of work on logic around 'merging' auth methods . 

This was motivated by wanting to be able to do things like :

 - Create a mission for people who haven't signed in. Eg this box is for {name: "sweet old lady", phone : "112-122-1343" } (ie recipient as user object without recipient having to login to our app) 
 - Allow said 'sweet old lady' to login, later, and auth with her phone number and 'claim' her existing user record at that time
- Allowing people to use their phone to authenticate the first time but then to avoid having to do the validate-sms-code-dance every time they login let them add a social media auth method to their account
- Allow people to sign up using their social media profile, and not put any obstacles in their way of their enthusiasm, let them have a look around and only once they actually get approved and want to accept missions do we hit them with "hey you need to add a phone auth to your account before you can actually accept a mission". (they might get excited and just donate money instead, you never know, and we don't need to make people give us their phone # in order to give us money or request missions for themselves or other people etc)

--

The first two use cases got moved to MVP1 but the idea was that we could build a general-purpose the system that let people authenticate in multiple ways to the same account. For this to work missions and what not need to be attached to our single user object instead of the actual auth record, though. 

--

Anyway the truth is I don't know what @mat10tng  ended up building exactly but I do know he spent ages getting it to all work and he said at one point that the firebase docs don't actually reflect how things work.. so just want to note that here in case that explains what is going on here..

src/app/model/User.ts
mat10tng: @utunga this is a good question, so missions completed in our book actually mean, a mission is completed only once the mission succeeds or failed, but where does mission delivered go for the volunteer view.

They only have, available, in progress, and completed

src/app/page/Home/Home.js
mat10tng: ehm, this is for the user who is logged in without going through the signup, to make sure we always have a profile to go with the system

src/app/component/AddressInput/AddressInput.jsx
mat10tng: hahahaha hahahaahaha what the ?

mat10tng:
> This is a monstrosity......
> User.test.ts was just completely broken and trying to call methods that didn't exist. Just got rid of it for now.
> 
> I fixed the weird UI issues with address field in the user profile.
> Fixed some things that changed with the payment flow.
> 
> I was looking back at the user docs for react-redux-firebase and I think we're saving our users wrong.
> 
> See here: http://react-redux-firebase.com/docs/recipes/profile.html#update-profile
> I think we need to be using this instead of just copying whatever is in `auth` into a user object in the db. Or is it the same thing?

Yes, profile and firestore.users are the samething
Might as well use the firestore users to clarify it was my thought

src/app/model/User.ts
utunga: Yes absolutely we should include both delivered and succeeded in the schema and in the way the app works. It's actually a really important step that a mission has to be not only marked as 'delivered' by the volunteer but actually accepted by the recipient. 

Not only did they drop the box off but the recipient was able to receive and find the box. 

Missions that are completed but not yet succeeded should be in the *IN PROGRESS* tab until they are fully accepted. If the organizer is looking at the In progress tab they should be watching to see that all boxes get not only **delivered** but also **accepted**. They may need to call up recipient or volunteer to work out what went wrong if not. 

In that case, they can override and just mark it as succeeded or failed. This is documented on the [Mission Flow diagram](https://www.figma.com/file/GLg0zPNnwMEGIb9TytCrxQ/Resilience-App?node-id=1257%3A601) _(FWIW I wish I could rename this diagram because its not just for Organisers but everyone)._ 

Its a pretty critical diagram and hasn't changed in a while by the way.



image

utunga: Hmm the swim lanes from top to bottom are recipient, volunteer, recipient.

The white boxes are actions you can undertake.

The gray boxes indicate MissionStatus, and actions between the dotted lines should only happen in that status.

PR #413, opened 2020-05-03
Work in progress -- Add components to communicate with routing on map using backend

CSumm:
Components will eventually communicate with [backend](https://github.com/factn/communitytech-tsp-backend
) developed by data science team that takes information and calculates shortest paths to routes and back
All components were taken from the [client side example that communicates with the backend ](https://github.com/factn/communitytech-tsp-frontend)mentioned above.

TODO: Get components into the correct page and test backend functionality for results


jwu910:
What ticket is this for?

CSumm:
This is for story 172 and 173 on Airtable; I am not sure if it has a ticket yet assigned

mat10tng:
can you check if dashboard work at all with your pr ? I have not take it down yet, but I suspect that the entire dashboard using map from react leaftlet

CSumm:
Dashboard is the view for volunteers, right?

CSumm:
Dashboard seems to work now, I think we're no longer using leaflet according to @utunga ; instead I'm feeding data from backend into google maps in a new window

CSumm:
@mat10tng closing this PR, as it will be integrated with the new dashboard (this one refers to the old one, and is not compatible with recent changes)

PR #421, opened 2020-05-04
now with npm run dev, we start a localdatabase instead

mat10tng:


firebase/database.rules.json
```diff
@@ -0,0 +1,6 @@
+{
+  "rules": {
```
qdozaq, line 2: We should copy the rules currently in the database. Though it may need a different format

package.json
```diff
@@ -57,7 +57,11 @@
     "start": "react-scripts start",
     "storybook": "start-storybook -p 9009 -s public",
     "test": "react-scripts test",
-    "upload": "python scheme/data.py && NODE_ENV=development node scheme/uploadMock.js"
+    "upload": "python scheme/data.py && NODE_ENV=development node scheme/uploadMock.js",
+    "dev": "npm-run-all --parallel dev:firebase dev:react dev:upload",
+    "dev:firebase": "firebase emulators:start",
```
qdozaq, line 62: we should run the `tsc` build for the functions folder here as well

jwu910:
> as well as instructions on installing pip and faker

Amy had a PR up for the docs. I may or may-not have merged it yet. but there were install instructions for python pip and faker

package.json
```diff
@@ -57,7 +57,12 @@
     "start": "react-scripts start",
     "storybook": "start-storybook -p 9009 -s public",
     "test": "react-scripts test",
-    "upload": "python scheme/data.py && NODE_ENV=development node scheme/uploadMock.js"
+    "upload": "NODE_ENV=development node scheme/uploadMock.js",
+    "dev": "REACT_APP_USE_DB_EMULATORS=true npm-run-all --sequential dev:functions_build  --parallel dev:firebase dev:react dev:upload",
+    "dev:functions_build": "cd functions && npm run build && cd ..",
+    "dev:firebase": "firebase emulators:start",
+    "dev:react": "react-scripts start",
```
jwu910, line 64: Is this a duplicate for `start`?

scheme/.gitignore
```diff
@@ -1 +1 @@
-*.json
\ No newline at end of file
+secret.json
```
jwu910, line 1: I thought we got rid of this? lol

package.json
mat10tng: yeah, but it is subject to change in case of environment, so its better with duplicated

mat10tng:
I removed out faker and pip dependecies for running `npm run dev` so, if you are not admin then you don't need it.... because that is only for generate data.json from data.py. This data.json would then populate mock data for local database, and stage data every two hours.

PR #453, opened 2020-05-04
Display and calculate route between pick up and delivery location in Google Maps

CSumm:
On view Route button click on 'Missions Group', it passes mission delivery and pickup location into url and opens Google Maps in a new window.

### DIFF
--- a/src/app/component/MissionGroup/MissionGroup.jsx
+++ b/src/app/component/MissionGroup/MissionGroup.jsx
@@ -55,6 +55,14 @@ const MissionGroup = ({ callToAction, group, groupCallToAction, showViewRoute })
   };
   const onClickViewRouteButton = () => {
     console.log("View Routes");
+    missions.forEach((mission) => {
+      console.log(mission);
+      let pickUpLocation = mission.pickUpLocation.address;
+      let deliveryLocation = mission.deliveryLocation.address;
+      window.open(
+        `https://www.google.com/maps/dir/?api=1&origin=${pickUpLocation}&destination=${deliveryLocation}&travelmode=driving`
+      );
+    });
   };
 
   const viewRoute = showViewRoute && (
--- a/src/app/component/MissionGroup/MissionGroup.jsx
+++ b/src/app/component/MissionGroup/MissionGroup.jsx
@@ -1,7 +1,7 @@
 import PropTypes from "prop-types";
 import React from "react";
 import clsx from "clsx";
-import { isLoaded } from "react-redux-firebase";
+import { isEmpty, isLoaded } from "react-redux-firebase";
 import { useHistory } from "react-router-dom";
 import MuiExpansionPanel from "@material-ui/core/ExpansionPanel";
 import MuiExpansionPanelActions from "@material-ui/core/ExpansionPanelActions";
@@ -11,6 +11,7 @@ import MuiExpandMoreIcon from "@material-ui/icons/ExpandMore";
 import MuiGrid from "@material-ui/core/Grid";
 import MuiMapIcon from "@material-ui/icons/Map";
 import { makeStyles } from "@material-ui/core/styles";
+import { ShowDeliveryRoute } from "../../component/";
 
 import { Button, MissionList } from "../index";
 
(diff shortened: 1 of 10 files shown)


src/app/component/MissionGroup/MissionGroup.jsx
```diff
@@ -55,6 +55,14 @@ const MissionGroup = ({ callToAction, group, groupCallToAction, showViewRoute })
   };
   const onClickViewRouteButton = () => {
     console.log("View Routes");
+    missions.forEach((mission) => {
+      console.log(mission);
```
jwu910, line 59: We need to remove debugging console logs

src/app/component/MissionGroup/MissionGroup.jsx
```diff
@@ -55,6 +55,14 @@ const MissionGroup = ({ callToAction, group, groupCallToAction, showViewRoute })
...
+      let pickUpLocation = mission.pickUpLocation.address;
+      let deliveryLocation = mission.deliveryLocation.address;
```
jwu910, line 61: We probably shouldn't use `let` for declarations if we're not reassigning these variables in this block scope.

src/app/component/MissionGroup/MissionGroup.jsx
```diff
@@ -55,6 +55,14 @@ const MissionGroup = ({ callToAction, group, groupCallToAction, showViewRoute })
...
+      window.open(
```
jwu910, line 62: Is it possible to generate an anchor to indicate a page navigation other than a window.open? I dont think this will play well with accessibility and or just web dev practices. I think it's pretty uncommon to invoke window methods on the window object this directly

src/app/component/MissionGroup/MissionGroup.jsx
```diff
@@ -55,6 +55,14 @@ const MissionGroup = ({ callToAction, group, groupCallToAction, showViewRoute })
   };
   const onClickViewRouteButton = () => {
     console.log("View Routes");
```
jwu910, line 57: If we're implementing this functionality on this callback, we're gonna need to remove this console log

src/app/component/MissionGroup/MissionGroup.jsx
CSumm: I was told to use the server instead of connecting to Google Maps directly. Trying to figure that out right now.

jwu910:
What ticket is this for, btw?

src/app/component/MissionGroup/MissionGroup.jsx
utunga: To be clear we wanna do this

1. Get the list of missions (and their addresses)
2. Send them to the server (API) and get back the optimum *ordering* of addresses
3. Take those addresses, in order, and send them to Google Maps directly as you are doing here.

utunga: @jwu910 I understand that generally we want to have the user explicitly click the link but the thing is it takes time to generate the link - like 2-3 seconds - so we don't want to have that happen when this component renders, only after they click the button - at which point we need to pop open some sort of interstitial or something - and *then * navigate the use to the URL we just generated (which will open google maps).

In such a case would use of window.open be acceptable? Or.. would you still think it's  better in your opinion to generate ana actual anchor `<a href></a>` and then simulate the 'click' event on that anchor?

utunga:
> What ticket is this for, btw?

It's for :
[172: "As a volunteer I can plan the route of any upcoming missions for efficiency (backend thing)"](https://airtable.com/tblRjd2McKgRW9MKj/viwlRRNI0PZIxrGkc/rec9K3kAHfLq6kdiI?blocks=hide)

And also (closely related) :
[173: "As a volunteer I can use the planned route of a mission and reference it for driving directions during a mission (google maps)"](https://airtable.com/tblRjd2McKgRW9MKj/viwlRRNI0PZIxrGkc/recSrbuy9DByyJru8?blocks=hide)

src/app/component/MissionGroup/MissionGroup.jsx
jwu910: Saw some changes to the code, didn't get a chance tor review yet, but wanted to comment on this.

As soon as the app knows about the stops, I think we want to generate that link. Even if its async, we should be able to run it as soon as the user is done with input.

The button can be disabled while we wait for teh async call, and then fill in the href src as soon as the results are back.

Thoughts?

utunga: I like the idea of an async loading of that button - minimizes the time it will take to get to the route .. the only thing i worry about is that a volunteer in a hurry may become really frustrated by a button that  ... won't.. f**ng ...   click... 

I think the solution is to yes use an async thing to load the button but somehow do something that shows the user clearly that we are calculating the optimal route as to why the button is disabled... 

of course in MVP1 territory we may just have to store the address order in the group or something..

jwu910: This is react and we have the incremental dom!

We can easily just render a grey button that says loading route or something and then it can change colors when it’s ready.

If we’re having issues with the time it takes to come back from the API, I think we’d need to figure something out then. 

But with the features available to us with React, we have quite a bit of manipulative freedom

utunga: OK cool.. thanks @jwu910 .. FWIW one of the reasons it takes a while is just that traveling salesperson problem is np complete so it may be that its actually not possible to make it go faster - that said I've suggested one thing to Vinh and he said he was gonna look into that. (I suggested that we hand his algorithm lat/lngs instead of asking his back end to do the geo-lookups)

src/app/component/MissionGroup/MissionGroup.jsx
utunga: No longer there

src/app/component/MissionGroup/MissionGroup.jsx
utunga: also no longer there

src/app/component/MissionGroup/MissionGroup.jsx
utunga: Implemented as Async using useEffect

utunga:
Closing this one for now so we can have a cleaner/easier to understand PR

PR #454, opened 2020-05-04
Organizer edit mission details - WIP (needs finishing or replacing) 

Eihcir0:
@mat10tng asked me to submit this wip pr because I will be busy the next couple days and he said he might have time to look at it.  

Here are the biggest issues remaining on this:

1. **AlgoliaPlaces** does not have a clean way to set an initial value -- they're recommended way is by directly manipulating the input node's `value` (like `document.querySelector("#outer-div-id input").value = myAddress`).  It would probably be cleanest to just create a wrapper around this that can take in `initialValue` and deal with it itself
2. The logic for the datetime selectors currently found in `MissionCreate` is pretty involved and highly coupled with the rendering component.  Probably best bet for that is to create a nice clean fully encapsulated wrapper component that just takes `value`, `onChange`, and `initialValue` and handles the rest.  Pull out all of that logic from MissionCreate
3.  I haven't even started looking at the selector for the volunteers  / phone number
4. Cleanup, this is still very much in a wip state.  The MissionEditView was copied from the DetailsView and I've just been focused on getting it working so sorry for the current state


mat10tng:
the work keep going in #619

PR #461, opened 2020-05-05
fix #459-type-string-for-fundedDate

### LINKED ISSUE #459: Type string does not match type .. for key fundedDate
Running localhost/master branch I was on this page:
image

And chose 'funded by donation' then it did this...
image

Just old data? Or ?

mat10tng:


utunga:
Thanks @mat10tng  !

PR #463, opened 2020-05-05
fix#460 where user profile page do not save

mat10tng:


src/app/model/User.ts
```diff
@@ -57,14 +57,20 @@ class User extends BaseModel {
...
    * @param {string} userUid - user
    * @param {object} data- updated data
    */
-  createProfile(userUid: string, data: object) {
-    return this.getCollection("users")
-      .doc(userUid)
-      .set(this.load({ data, uid: userUid }));
+  createProfileIfNotExist(auth: any, profile: any) {
+    if (!auth || !profile) return;
+    if (auth.isLoaded && !auth.isEmpty && profile.isEmpty && profile.isLoaded) {
```
qdozaq, line 71: I'm not sure this works.
I logged out my auth object and profile object
image

profile still comes through as not empty so this would never get triggered

qdozaq: Also could we make this a seperate function?

Theres still some other pages that rely on the original createProfile function

jwu910:
I used a user context provider to manage firebase user profiles before.

wan't me to take a look at this?

src/app/model/User.ts
mat10tng: ehm, this is for create profile for the very first time, you already have your profile created, if you want to stimulate this, you need to either
1. with simulator, just restart it
2. on stage, remove your profile/user from firestore

yes, I will create both function

mat10tng:
> I used a user context provider to manage firebase user profiles before.
> 
> wan't me to take a look at this?

How does it mange user profiles?

jwu910:
> How does it mange user profiles?

What I did before, was on login, i wrote to a user context the info i needed from the firebase user auth object.

Upon logout, we'd clear that user object in context, and then triggered a redirect back to home, or out of the protected route.

The plan was to be able to just reference the user context if we needed to get information about the user

mat10tng:
Ok, so the user context is a wrapper that make sure we using auth -> currentUser ? The problem here is that there are two way of accessing the dashboard
1. First you sign up, then you login -> this is awesome, good, then we have initialize the user in firestore properly
2. You just sign in, using the signing button, without signup -> ok, we have a problem because suddenly we have some one that have auth in the firebase without a profile, ( represent by being a user document in firestore.users ). So, I solved it by check this specifict problem, where, we always create a profile if we notice it is empty from the auth system.

PR #475, opened 2020-05-07
Optimise typography and font style

### LINKED ISSUE #276: Optimization: Typography and Font styles are written many times almost exactly the same
Typography and font styles is an area to optimize some code that seems to be written many times (wet).

Some optimizations can be made as referenced below:

https://github.com/factn/resilience-app/pull/267#pullrequestreview-395928293

https://github.com/factn/resilience-app/pull/267#discussion_r410718719

marekdano:
I created more typography components and replace Typography-s with the typography component across the whole project.

I added a unit test for the MissionCard component and planning to add more tests for other components.

Closes #276


src/app/component/Typography.jsx
```diff
@@ -26,6 +28,11 @@ const H5 = ({ children, ...rest }) => (
...
   </Typography>
 );
+const H6 = ({ children, ...rest }) => (
+  <Typography variant="h6" {...rest}>
+    {children}{" "}
+  </Typography>
+);
 const Body1 = ({ children, ...rest }) => (
   <Typography variant="body1" {...rest}>
     {children}{" "}
```
mat10tng, line 38: extra space

marekdano: It was there. Not sure why we have it. Do you think that we should remove it?

mat10tng: yes, we should

marekdano:
I have to fix fail tests on MissionCard before this PR is merged.

PR #484, opened 2020-05-08
add map to address input, enhance disabled state

### LINKED ISSUE #471: We need a better address picker
In this [comment on a pr](https://github.com/factn/resilience-app/issues/387#issuecomment-623263341) we talked about the three places we use maps in the app.

Specifically
- Organizer dashboard
- Volunteer dashboard (mostly trying to defer to Google Maps)
- Address pickers (for user profile and setting mission pickup/drop off locations). 

Last night, about 1 am, in trying to get the google maps thing to work I had a sinking realization that we have a pretty major usability problem with the address picker.

We have the Algolia address picker, but I am concerned that as currently, the Algolia method for typing and geolocating addresses is going to have a lot of problems. I suggest you go ahead and try it on the [Request Foodbox](https://resilience-app.herokuapp.com/request/foodbox) interface.

The UI looks sweet by the way! But if you try to enter real actual addresses you encounter lots of problems (or at least I did). I'm worried we're going to end up with lots of badly entered addresses and (worse) people not being able to correctly pick up or deliver parcels. 

I already filed #468 about the specific problem of not being able to enter apartment number but I think we're going to have worse problems than that. 

For instance, if trying to enter "Unit 3, 123 Coldwater Canyon,  CA 91423" the Algolia picker does this:

image

By contrast, the google picker is not much better - it does this:
image

In either case, the end-user needs to know that they need to reformat their address until it is actually parsed correctly by the geolocation API. 

Like this:
image

But getting them to actually do this correctly is vital. Firstly, so that our volunteers don't get confused. And secondly, we need that nicely formed address so that - when we fire the address back at Google Maps for the turn by turn directions - Google Maps is actually able to actually render the location/route. 

--

I did a bit of research on this and from a UX perspective, the Uber location picker is considered one of the best out there. What's super cool is, after you enter the address, you can then use zoom and pan to drag the pin to put the pin in the exact right location..

image

Being able to fine-tune like Uber is probably an MVP1 type feature but as an interim feature, I think we really need to pair these address fields with a map just under, or above the address, so that they can get a preview of where they mean.

That is, we don't want to just show them that Algolia or Googlemaps was *able* to parse the address but that it was able to parse it to *this specific address* - thus giving the user time to correct it if its gonna be in the wrong place. 

My current thought is we have something like:
+-------------+
|      map        |
+-------------+
[address input]
[address detail]

So they can update the address input and the map preview updates in real time, and then directly under that a second input for putting stuff like 'Unit 2' or '(NOHO Police Station)' Or whatever. That would solve  #468 at the same time.

What do folks think? Can we do a general-purpose component for this? Maybe react-leaflet-location-picker is all we need? Or is it just a good start?

Thanks! 

--

PS
The address detail field would be for stuff specifically relating to the address (such as "Unit 2"). Other directions (such as "please press the buzzer and wait for the super, the dog has COVID and Rabies so whatever you do don't open the gate' ) would go into the actual mission notes....

mat10tng:
Fixes #471 

https://tinyurl.com/yblzvq9c


utunga:
Looks absolutely gorgeous ! There seems to be a bug? Or problem in the way that the actual geo-location is working however. Doesnt seem to be using the street address number?

If I copy and paste Unit 3, 123 Coldwater Canyon, CA 91423 into the address bar I get this..

image

If I then realise no thats not 123 Coldwater Canyon and manually fix it, I get this (notice that the actual location pin hasn't moved)
image

In my testing only changing the actual street made the pin move didn't seem to pay attention to the street number at all, for any addresses I was able to try...

src/app/component/AddressAutocomplete.jsx
```diff
@@ -0,0 +1,295 @@
...
+  };
+
+  const fetchLatLng = useMemo(
+    () =>
+      throttle((request, callback) => {
+        geocoderService.current.geocode(request, callback);
+      }, 200),
+    []
+  );
+  const fetch = useMemo(
```
qdozaq, line 173: Might want to use a different name here so it's not confused with global `fetch`

package.json
```diff
@@ -12,6 +12,7 @@
     "@material-ui/pickers": "^3.2.10",
     "@reduxjs/toolkit": "^1.3.4",
     "algolia-places-react": "^1.5.1",
```
Eihcir0, line 14: There's still one naked reference to Algolia -- in `MissionForm` (for mission create) which should be dealt with (replaced with *AddressInput*) either in this pr or a separate.  Actually,  I have a [pr in on that form ](https://github.com/factn/resilience-app/pull/492)to switch it to the new DateTimeInput... I could expand the scope of that pr and get us off Algolia altogether?? @mat10tng @qdozaq

src/app/component/AddressInput/AddressInput.jsx
```diff
@@ -25,27 +46,79 @@ const AddressInput = (props) => {
...
+      }
+    }
+  }, [location, zoom, position]);
+
+  const marker = location ? (
+    <Marker
+      position={location}
+      draggable={true}
+      riseOnHover={true}
+      onMoveend={(e) => {
```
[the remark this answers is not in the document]
mat10tng: yepp, but its not working properly with the map anyhow, I will fix it

package.json
mat10tng: yeah, lets make another issue for getting rid of address input together with Algolia instead

src/app/utils/loadScipts.js
```diff
@@ -0,0 +1,33 @@
+/**
+ *
+ * @param {string} src : src of the script
+ * @param {string} position : position of the script in the document
+ * @param {script id} id: the unique id of the script
+ */
+export function loadScript(src, position, id) {
```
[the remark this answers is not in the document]
mat10tng: done

src/app/component/AddressInput/AddressInput.jsx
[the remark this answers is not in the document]
mat10tng: done

src/app/component/AddressAutocomplete.jsx
mat10tng: done

package.json
mat10tng: https://github.com/factn/resilience-app/issues/494

src/app/hooks/useGoogleMapAPI.js
```diff
@@ -0,0 +1,36 @@
...
+ * useGoogleMapApi
+ *
+ * @returns {false | true} google map api is loaded
+ */
+export default function useGoogleMapApi() {
+  const [loaded, setLoaded] = useState(false);
+
+  useEffect(() => {
+    // Check if paypal has already been loaded
+    if (document.querySelector("#google-maps")) {
```
qdozaq, line 14: Would it be useful to pass in this id into the hook?

qdozaq: Also the comment lol

mat10tng: not unless we decide to make a hook for useScript instead....
let me remove that comment hahahahah

mat10tng: I mean, we could do, in the future
useScripts: id, src

PR #488, opened 2020-05-08
Introduce snackbar context

### LINKED ISSUE #486: Create Snackbar context
Component should be wrapped in a snackbar context provider with a callback to update snackbar

jwu910:
Fixes #486 <br /> Added snackbar context and used the api in a couple components. Wont have time to apply this change across all of the other snackbar implementations right now, but maybe we can create a new ticket to update the rest of snackbar usage later

Usage: 
```
import {SnackbarContext} from '..../component/Snackbars/Context';

const snackbarContext = useContext(SnackbarContext);

		// default type is success
      snackbarContext.updateSnackbar({
        message: `Some meaningful error message`,
        type: "error",
      });
```

### DIFF
--- a/src/app/page/MissionCreate/MissionCreate.jsx
+++ b/src/app/page/MissionCreate/MissionCreate.jsx
@@ -1,10 +1,10 @@
 import CircularProgress from "@material-ui/core/CircularProgress";
-import React, { useState } from "react";
+import React, { useContext, useState } from "react";
 import { getFirebase, withFirestore } from "react-redux-firebase";
 import { withRouter } from "react-router-dom";
 
-import { ErrorSnackbar, SuccessSnackbar } from "../../component/Snackbars";
 import useForm from "../../hooks/useForm";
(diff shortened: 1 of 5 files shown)


src/app/page/MissionCreate/MissionCreate.jsx
```diff
@@ -76,17 +74,26 @@ function MakeMission({ history }) {
       Mission.create(mission)
         .then(() => {
           setLoading(false);
-          setSuccessSnackbarOpen(true);
+
+          snackbarContext.updateSnackbar({
```
mat10tng, line 78: maybe instead of having updateSnackbar, we could just have snackbarContext.show(....) ?

jwu910: I can update the method name

jwu910: Done :+1:

PR #491, opened 2020-05-08
Date time input

Eihcir0:
This pr creates new `DateTimeInput` component which wraps _KeyboardTimePicker_ (material-ui) and our _KeyDatePickerContainer_ -- a wrapper around _KeyboardDatePicker_ (material-ui)

I needed this in Mission edit and I saw that we were using this logic in MissionForm so I created this reusable component to be used by both.  

The next pr (dependent upon this one) plugs this component into the MissionForm (create) and then a subsequent pr after that will see this used in the MissionEditForm


jwu910:
Do you have an airtable or github issue number I can attach this to?

mat10tng:
no, this unfortunately was only word of mouth between me and Richie

Issue #435, opened 2020-05-04
Donors can donate money online to fund boxes


DAHuerta:
**Description**

As a donor, I can donate money online to fund food boxes (to be assigned by the organizer) using online payment and get a receipt with charitable status.

[Airtable #143](https://airtable.com/tblRjd2McKgRW9MKj/viwlRRNI0PZIxrGkc/recYoIuMMeHo45Iag?blocks=hide)

[Figma](https://www.figma.com/file/GLg0zPNnwMEGIb9TytCrxQ/Resilience-App?node-id=29305%3A10722)

PR #493, opened 2020-05-09
Feat/143 donations

qdozaq:
closes https://github.com/factn/resilience-app/issues/435

image

image

image


firebase/firestore.rules
```diff
@@ -19,6 +19,10 @@ service cloud.firestore {
       	allow read;
       	allow write: if isOrganizerOf(organizationId);
       }
+      match /donations/{recieptid} {
+        allow read: if isOrganizerOf(organizationId);
+        allow write;
+      }
```
qdozaq, line 25: I updated this in firebase as well

src/app/component/PaypalCheckout/PaypalCheckout.tsx
```diff
@@ -1,27 +1,38 @@
...
+  ...rest
+}: Props) {
   const paypal: any = usePaypal();
 
   function createOrder(data: any, actions: any) {
     return actions.order.create({
       purchase_units: Array.isArray(cart) ? cart : [cart],
       application_context: {
         shipping_preference: "NO_SHIPPING",
+        brand_name: Organization.data.name,
```
qdozaq, line 35: This isn't doing anything right now. Will update when I make the org context

src/app/model/Organization.ts
```diff
@@ -74,6 +74,16 @@ class Organization extends BaseModel {
       return { clientUid: "sb", email: "" };
     }
   }
+
+  logDonation(donation: DonationLog) {
```
qdozaq, line 78: As a just in case I'm logging donations. We don't use them for anything but better to track in the event we need them

src/app/page/Donate/DonationPage.jsx
```diff
@@ -0,0 +1,187 @@
...
+    ...theme.typography.body1,
+    display: "flex",
+    flexDirection: "column",
+    alignItems: "start",
+  },
+}));
+
+const orgPhoneNumber = "555-555-5555";
+const organizationEIN = "XX-XXXXXXX";
+const organizationName = "Org_name_replace_me";
```
qdozaq, line 42: Will be replaced with the org context

src/app/page/Donate/DonationPage.jsx
```diff
@@ -0,0 +1,187 @@
...
+  );
+}
+
+function formatPaypalDetails(details) {
+  const { payer, purchase_units } = details;
+  const donation = purchase_units[0];
+
+  return {
+    amount: donation?.amount?.value,
+    createdDate: Date.now(),
```
mat10tng, line 70: we really should add to the lodash -> _.getDate() to make sure its unified

PR #497, opened 2020-05-09
Implement organizer context

qdozaq:
Creating the organizer context and hook.

Think I got all the places that need org data. Let me know if I missed anything.


src/App.jsx
```diff
@@ -44,49 +46,70 @@ function PrivateRoute({ children, ...rest }) {
...
+// grab from domain or some service
+const ORGANIZATION_ID = "1";
+
 function App() {
+  const [org, setOrg] = useState();
+  useEffect(() => {
+    Organization.init(ORGANIZATION_ID)
+      .then((org) => setOrg(org))
+      .catch((error) => {
+        console.error(error);
```
mat10tng, line 58: should we do a snackbar error here ?

src/App.jsx
```diff
@@ -44,49 +46,70 @@ function PrivateRoute({ children, ...rest }) {
...
-              </Switch>
-              <Snackbar.Context.SnackbarConsumer>
-                {(value) => {
-                  return <Snackbar handleClose={value.closeSnackbar} {...value.snackbar} />;
-                }}
-              </Snackbar.Context.SnackbarConsumer>
-            </Snackbar.Context.SnackbarProvider>
+            <OrganizationContext.Provider value={org}>
+              <Snackbar.Context.SnackbarProvider>
+                {org === null && <Redirect to="/404" />}
```
mat10tng, line 75: I think there is some new way of handle routing, by Amy

src/App.jsx
qdozaq: Hmm there's not currently a good way to set the snackbar while outside of the snackbar provider. Something to look at later

jwu910: Depends on how the app should respond to the app unable to run the init org method right?

If we’re dependent on these org hook being set in the provider, maybe we need a more serious error lol. 

Will the app work if the org is not set?

qdozaq: Everything basically depends on the org being set in order to make certain calls. So right now I just have it redirecting to the organizer signup page if it doesn't exist

PR #498, opened 2020-05-09
Add error boundary component and use it in two 

### LINKED ISSUE #485: CSVLink fail shouldn't be able to take down the whole dashboard page
Since CSVLink currently renders all data to CSV inside of the button, if something goes wrong with it, it takes down the whole entire dashboard - renders as white.

So apart from actually fixing the bug I /we also need to make sure it doesn't take down the dashboard home if it breaks.

image

utunga:
... so that we dont bring down whole entire app if something breaks.

Fixes #485

### DIFF
--- a/src/app/component/DownloadCsv/DownloadMissionsCsv.jsx
+++ b/src/app/component/DownloadCsv/DownloadMissionsCsv.jsx
@@ -3,6 +3,7 @@ import dot from "dot-object";
 import PropTypes from "prop-types";
 import React from "react";
 import _ from "lodash";
+import ErrorBoundary from "../ErrorBoundary";
 import Grid from "@material-ui/core/Grid";
 
 const DownloadMissionsCsv = ({ data, filename, label }) => {
@@ -28,14 +29,16 @@ const DownloadMissionsCsv = ({ data, filename, label }) => {
   //const headers = {};
   return (
     <Grid item>
-      <CSVLink
-        data={flattened}
-        filename={fileNameWithTimestamp}
-        className="btn btn-primary"
-        target="_blank"
-      >
-        {label}
-      </CSVLink>
+      <ErrorBoundary>
+        <CSVLink
+          data={flattened}
+          filename={fileNameWithTimestamp}
+          className="btn btn-primary"
+          target="_blank"
+        >
+          {label}
+        </CSVLink>
+      </ErrorBoundary>
     </Grid>
   );
 };
--- a/src/app/component/ErrorBoundary.js
+++ b/src/app/component/ErrorBoundary.js
@@ -1,0 +1,31 @@
+import React from "react";
+
+class ErrorBoundary extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = { hasError: false };
+  }
+
+  static getDerivedStateFromError(error) {
+    // Update state so the next render will show the fallback UI.
+    return { hasError: true };
+  }
+
+  componentDidCatch(error, errorInfo) {
+    // You can also log the error to an error reporting service
+    console.error(error, errorInfo);
+    //
+    // TODO hook up to raygun or similar
+    // logErrorToService(error, errorINfo)
+  }
+
+  render() {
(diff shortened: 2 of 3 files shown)


src/app/component/ShowDeliveryRoute/ShowDeliveryRoute.jsx
```diff
@@ -89,11 +90,13 @@ const ShowDeliveryRoute = ({ missions }) => {
   }
 
   return (
-    <Link href={googleMapsUrl} target="_blank">
-      <Button fullWidth={true} variant="text" startIcon={<MuiMapIcon />}>
-        View Route
-      </Button>
-    </Link>
+    <ErrorBoundary>
+      <Link href={googleMapsUrl} target="_blank">
```
qdozaq, line 94: Was this erroring because of the the external googMapsUrl?

I don't believe `Link` can accept urls that dont point to the site. We should use anchor tag here

src/app/component/DownloadCsv/DownloadMissionsCsv.jsx
```diff
@@ -28,14 +29,16 @@ const DownloadMissionsCsv = ({ data, filename, label }) => {
...
-      <CSVLink
-        data={flattened}
-        filename={fileNameWithTimestamp}
-        className="btn btn-primary"
-        target="_blank"
-      >
-        {label}
-      </CSVLink>
+      <ErrorBoundary>
+        <CSVLink
```
qdozaq, line 33: Do we know what the problem is yet with `CSVLink`?

utunga: Yes we do. But the thing is no matter what the broken data problem is, it shouldn't take down the whole entire dashboard. 

So, my thought was to start by fixing the symptom then will look at fixing the data making it so that _.flatten(_.map doesn't get thrown by inconsistent 'dummy' data./

src/app/component/ShowDeliveryRoute/ShowDeliveryRoute.jsx
utunga: It wasn't erroring at all but because it uses an external API (the python back end) it seems like an appropriate place for an ErrorBoundary.. if there is something wrong with the python back end it shouldn't take down the app, just the ability to view optimized routes in google maps.

utunga: fwiw the 
```
 <Link href={googleMapsUrl}
```

Was from master (earlier approved PR). It 'works' like this. Should I still change it to an anchor tag?

jwu910: I think external URLs should be regular anchors right? <a href /> rather than Link--part of the react router api.

Also any time we use target="_blank" We should include a "noopener noreferrer" in a rel attribute to protect from i think tab injection. I forgot the immediate risk that new tabs expose, I'll have to look it up later when i get a minute

src/app/component/ErrorBoundary.js
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  render() {
+    if (this.state.hasError) {
+      // You can render any custom fallback UI
+      return "";
+    }
+    return this.props.children;
+  }
+}
```
jwu910, line 29: Do we want to render nothing if there is an error? My I think the implementation looks fine, just concerned that a blank space in place where the component is supposed to be might not be indicative that an error was thrown. Do we want to keep the error message and display that here? like a "oops something isn't working, please refresh" message?

src/app/component/DownloadCsv/DownloadMissionsCsv.jsx
thirdeyeclub: CVS is indeed what is causing the breaking error.

utunga:
Merged as it resolves a blocker. Going to address @jwu910 's feedback in future PRs.

PR #501, opened 2020-05-10
Fix the main style of font-family 

marekdano:
I changed the font-family style group in the main component App. The font style works fine on my local branch now.

Please review and let me know if the font-family group should be different.

@amylo please review


amylo:
You should align with Typography described in our styleguide: https://www.figma.com/file/GLg0zPNnwMEGIb9TytCrxQ/Resilience-App?node-id=1256%3A1064

marekdano:
Each Typography components had `font-family: Open Sans` defined in. I removed it from all typography components and move it to `.App` style in `App.css` file. The issue with font styling has been introduced.
Checking the styleguide on Figma seems that each typography has to have `font-family: Open Sans` defined. Do you think that I should revert the change having `font-family` to be defined in each typography component instead of in the `.App`?

qdozaq:
@marekdano I believe the problem is you're trying to use global css to override the defaults in material ui, which doesn't work.

Instead you will need to override the theme object. Looks like you can set global font params at the root of the `typography` property https://material-ui.com/customization/typography/#font-family

PR #502, opened 2020-05-10
Feature/222 routes

amylo:
1. Created the Page Not Found "404" error page
2. Created the Unauthorized error page
3. Consolidated all the places we've hardcoded paths and moved them into a single file


src/app/page/Error/Unauthorized.jsx
```diff
@@ -0,0 +1,53 @@
...
+import { Body1, H1 } from "../../component/Typography";
+import unauthorizedImg from "../../../img/app-illustrations/error_401.png";
+import { routes } from "../../routing";
+
+const unauthorizedStyles = makeStyles((theme) => ({
+  graphic: {
+    height: "100%",
+    width: "100%",
+  },
+  heading: Object.assign(theme.typography.h1, {
```
mat10tng, line 16: this vs spread operator? just curious about the choice

src/app/page/Error/ErrorLanding.jsx
```diff
@@ -0,0 +1,41 @@
...
+ * Component for Error landing page
+ *
+ * @component
+ */
+const ErrorLanding = ({ errorCode }) => {
+  const classes = errorLandingStyles();
+
+  return (
+    <Page className="error-page">
+      <MuiContainer className={clsx("error-page-content", classes.errorContent)}>
```
mat10tng, line 26: are we moving toward a css based system as well ?

src/app/page/Error/Unauthorized.jsx
amylo: lol its old school way, since we can use ES6, I'll change it to spread

src/app/page/Error/ErrorLanding.jsx
amylo: i've just added it in case we need it, and also helps with semantics. The styling itself is all still done via React styles currently

PR #505, opened 2020-05-11
Feat/more realistic addresses

utunga:
Added data based on (anonymized versions of) real drop off addresses and real data. 

Since everything moved to LA switched up the dashboard home to use the org location to center it (whichi s down in LA instead of SFO).


mat10tng:
you need npm run format

PR #510, opened 2020-05-12
Fixing issues related to clicking action buttons on volunteer dashboard

laredotornado:


src/app/component/MissionComponent/MissionCard.jsx
```diff
@@ -129,38 +129,60 @@ const styles = (theme) => ({
  *
  * @component
  */
-const MissionCard = withStyles(styles)(({ anchorEl, classes, mission }) => {
+const MissionCard = withStyles(styles)(({ anchorEl, classes, handleUpdatedMissions, mission }) => {
+  console.log("handle updated missions ...");
```
mat10tng, line 133: console.log

src/app/page/Home/VolunteerHome.jsx
```diff
@@ -101,6 +102,17 @@ export default function VolunteerHome({ currentUser }) {
     updateInProgressMissions(inProgressMissions.filter((m) => m.uid !== missionUid));
   };
 
+  const handleUpdatedMissions = () => {
```
mat10tng, line 105: a better way of doing this would be, 
do the request in the button, and then refetch what need to refetch, to  explicit pass down this function to the component make the component depends on this props, and can not be reused

src/app/component/MissionComponent/MissionCard.jsx
laredotornado: Oops, removed.

src/app/page/Home/VolunteerHome.jsx
laredotornado: Ok.  The reason I did that was because after clicking those buttons it changes the state in the top level component (VolunteerHome) and I couldn't figure out how to get a reference to the VolunteerHome's states without passing the function down the button.  But I'll research this more and find a more elegant way.

src/app/component/MissionComponent/AcceptMissionButton.jsx
```diff
@@ -4,14 +4,16 @@ import CheckIcon from "@material-ui/icons/Check";
 import Mission from "../../model/Mission";
 
 /**
- * @param {userUID} the user UID that accept the mission
+ * @param {handleUpdatedMissions} function that updates missions tied to a user.
  * @param {user} the user that accept the mission
  * @param {mission} the mission to be accepted for
  */
-const AcceptMissionButton = ({ mission, user }) => {
+const AcceptMissionButton = ({ handleUpdatedMissions, mission, user }) => {
```
qdozaq, line 11: So this button along with `DeliverMissionButton` `StartMissionButton` and `UnassignMeButton` could all be the component.
Since you're not really styling the button in any special way you probably don't need these at all.
and really this also a good example for composition. 

See here
```js

({mission, ...}) => {

const acceptHandler = () => Mission.accept(mission, user, ...);

...

return <MissionCard>
   <Button {...someProps} onClick={acceptHandler}>
     Accept
   </Button>
   <Button {...someOtherProps} onClick={deliverHandler}>
     Deliver
   </Button>
...
</MissionCard>
}
```

src/app/page/Home/VolunteerHome.jsx
qdozaq, line 105: You may actually just be able to avoid all of this by using [useFirestoreConnect](https://react-redux-firebase.com/docs/firestore#useFirestoreConnect)
Which will auto sync all your data like we do in the org dashboard https://github.com/factn/resilience-app/blob/7bfc454f154d464a2d6e6162b0a4b02f2f29894d/src/app/dashboard/MissionsControl.jsx#L57

laredotornado: Ok, have updated to useFirestoreConnect .  Let me know what you think.

src/app/model/Mission.ts
```diff
@@ -138,6 +138,36 @@ const fsIncomplete = (orgId: string) => ({
   storeAs: "incompleteMissions",
 });
 
+const fsAvailableUserMissions = (orgId: string, userId: string) => ({
```
qdozaq, line 141: Can we just make this all one call and then filter it on the frontend?

I know this is how the org dashboard does it, but that should probably get refactored too.

This way we're cutting down how many queries (reads) are getting made.

src/app/page/Home/VolunteerHome.jsx
```diff
@@ -34,73 +39,42 @@ const useStyles = makeStyles((theme) => ({
...
+  console.log(missions[0]);
+
   const classes = useStyles();
   const [value, setValue] = useState(0);
   const [userPhoneUnverifiedPopupOpen, setUserPhoneUnverifiedPopupOpen] = useState(false);
-  const [acceptedMissions, updateAssignedMissions] = useState([]);
+  //const [acceptedMissions, updateAcceptedMissions] = useState([]);
   const [inProgressMissions, updateInProgressMissions] = useState([]);
-  const [availableMissions, updateAvailableMissions] = useState([]);
+  //const [availableMissions, updateAvailableMissions] = useState([]);
```
qdozaq, line 51: Can all these comments be removed?

src/app/component/MissionComponent/AcceptMissionButton.jsx
qdozaq: What were you're thoughts on this?

qdozaq: Also looks like we can remove the `handleUpdatedMissions` prop

src/app/model/Mission.ts
utunga: Can you add an issue to do that refactor later? Just being the usual PITA PO and saying hey we gotta get this out by Monday for reasl though. Promise we'll have a refactor party afterwards though!

src/app/component/MissionComponent/AcceptMissionButton.jsx
```diff
@@ -4,11 +4,11 @@ import CheckIcon from "@material-ui/icons/Check";
 import Mission from "../../model/Mission";
 
 /**
- * @param {userUID} the user UID that accept the mission
+ * @param {handleUpdatedMissions} function that updates missions tied to a user.
```
mat10tng, line 7: we dont use this function anymore ? Maybe we should do as Paul comment and remove this now that we have everything refractor to the model system

src/app/component/MissionComponent/AcceptMissionButton.jsx
mat10tng: yeah agree

src/app/model/Mission.ts
mat10tng: yeah, lets do that later.

src/app/page/Home/VolunteerHome.jsx
```diff
@@ -34,73 +39,22 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
-export default function VolunteerHome({ currentUser }) {
+const VolunteerHome = ({ currentUser, missions }) => {
+  console.log("\n\n\n==== missions ====");
```
mat10tng, line 43: console

laredotornado: Oops!  These are removed.

src/app/component/MissionComponent/AcceptMissionButton.jsx
laredotornado: Correct -- I removed it because I thought you said you wanted to have reusable components and so we moved the volunteer/mission-state tracking to the firebase redux store.

src/app/page/Home/VolunteerHome.jsx
laredotornado: Done!

PR #515, opened 2020-05-13
Production and Staging deployments

qdozaq:


.firebaserc
```diff
@@ -1,5 +1,6 @@
 {
   "projects": {
-    "default": "mutualaid-757f6"
+    "default": "mutualaid-757f6",
+    "production": "resilienceapp-6a0dd"
```
qdozaq, line 4: you can now use `firebase use <project>` to switch between the different databases and credentials

firebase.json
```diff
@@ -16,15 +16,29 @@
...
+  },
   "emulators": {
     "functions": {
       "port": 5001
     },
     "firestore": {
       "port": 8080
     },
-    "database": {
-      "port": 9000
```
qdozaq, line 27: This is fore realtime database which we dont use

firebase.json
```diff
@@ -16,15 +16,29 @@
...
+    "hosting": {
+      "port": 5000
```
qdozaq, line 41: port for local firebase hosting

package.json
```diff
@@ -132,6 +137,7 @@
     "eslint-plugin-sort-destructure-keys": "^1.3.4",
     "export-to-csv": "^0.2.1",
     "firebase-admin": "^8.12.1",
+    "google-gax": "^2.3.1",
```
qdozaq, line 140: I needed to add this to fix the weird credential issue I was having with `uploadMocks`

package.json
```diff
@@ -82,6 +84,9 @@
       "simple-import-sort",
       "sort-destructure-keys"
     ],
+    "globals": {
+      "firebase": "readonly"
```
qdozaq, line 88: firebase is a global now. See other comments for why

package.json
```diff
@@ -61,15 +61,17 @@
     "prettier:check": "prettier --check ./src",
     "prettier:fix": "prettier --write ./src",
     "server": "node server.js",
-    "start": "cross-env REACT_APP_USE_DB_EMULATORS=false react-scripts start",
+    "start": "run-p --race dev:firebase dev:react",
```
qdozaq, line 64: Still does the same thing. Only now it uses firebase to serve the react app. this is really cool cause now if we switch between production and staging on our local it will automatically switch to that DB. gonna make solving prod issues so much easier

public/index.html
```diff
@@ -35,14 +35,24 @@
...
-    ></script>
+    <!-- The core Firebase JS SDK is always required and must be listed first -->
+    <script src="/__/firebase/7.14.3/firebase-app.js"></script>
 
+    <!-- TODO: Add SDKs for Firebase products that you want to use
+     https://firebase.google.com/docs/web/setup#available-libraries -->
+    <script src="/__/firebase/7.14.3/firebase-analytics.js"></script>
+    <script src="/__/firebase/7.14.3/firebase-auth.js"></script>
+    <script src="/__/firebase/7.14.3/firebase-firestore.js"></script>
+    <script src="/__/firebase/7.14.3/firebase-storage.js"></script>
```
qdozaq, line 46: Switching the app to use firebase's hosted packages instead of the ones provided by npm.
This creates a global `firebase` variable on the window.

The reason for this is that it allows us to completely remove the firebase config and initializeApp calls as they're now auto configured by the firebase host, which means we can easily switch between as many different prod or staging environments as we want without having to worry about setting up the config for that project.

package.json
```diff
@@ -33,7 +34,6 @@
     "react-csv": "^2.0.3",
     "react-dom": "^16.13.1",
     "react-facebook-login": "^4.1.1",
-    "react-firebaseui": "^4.1.0",
```
qdozaq, line 36: See other comments for why this was removed

public/index.html
```diff
@@ -35,14 +35,24 @@
...
+
+    <!-- Initialize Firebase -->
+    <script src="/__/firebase/init.js"></script>
+    <script src="https://www.gstatic.com/firebasejs/ui/4.5.0/firebase-ui-auth.js"></script>
```
qdozaq, line 50: This script replaces the `react-firebaseui` module as the module doesn't play nice with the new script tag implementation of firebase

src/App.jsx
```diff
@@ -3,7 +3,6 @@ import "./App.css";
 import CssBaseline from "@material-ui/core/CssBaseline";
 import React, { useEffect, useState } from "react";
 import { BrowserRouter as Router, Redirect, Route, Switch } from "react-router-dom";
-import "firebase/storage";
```
qdozaq, line 6: cleanup.
this and other are now provided by the script tags

src/app/component/FirebaseAuthUi/FirebaseAuthUi.jsx
```diff
@@ -1,18 +1,27 @@
...
+let ui;
 export default function FirebaseAuthUi(props) {
-  const firebaseUiConfig = props.firebaseUiConfig || defaultFirebaseUiConfig;
-  return <StyledFirebaseAuth uiConfig={firebaseUiConfig} firebaseAuth={firebase.auth()} />;
+  const firebase = useFirebase();
+  const firebaseUiConfig = props.firebaseUiConfig || defaultFirebaseUiConfig(firebase);
+
+  useEffect(() => {
+    if (!ui) {
+      ui = new window.firebaseui.auth.AuthUI(firebase.auth());
```
qdozaq, line 21: Re-implementing this component with the new [firebaseui-web](https://github.com/firebase/firebaseui-web) script I commented on earlier

src/app/model/Organization.ts
```diff
@@ -4,7 +4,7 @@ import { Resource } from "./schema";
 import { createContext, useContext } from "react";
 
 const defaultData: OrganizationInterface = {
-  uid: "",
+  uid: "-1",
```
qdozaq, line 7: There was a weird error that would break the site if this was just an empty string. initializing this just incase

src/firebase.d.ts
```diff
@@ -0,0 +1,5 @@
+import firebase from "firebase";
+
+declare global {
+  export const firebase = firebase;
```
qdozaq, line 4: global declaration for typescript

src/index.tsx
```diff
@@ -1,24 +1,26 @@
...
 import { createFirestoreInstance } from "redux-firestore"; // <- needed if using firestore
-import { rfConfig } from "./config/firebase";
 
 import App from "./App";
 import { store } from "./app/store";
 import * as serviceWorker from "./serviceWorker";
 
-import initFirebase from "./initFirebase";
+const db = firebase.firestore();
+if (process.env.REACT_APP_USE_DB_EMULATORS === "true") {
```
qdozaq, line 12: connects to the emulated firestore

src/setupProxy.js
```diff
@@ -0,0 +1,9 @@
+const { createProxyMiddleware } = require("http-proxy-middleware");
+module.exports = function (app) {
+  app.use(
+    "/__",
```
qdozaq, line 4: The script tags I added all point to `/__`.
this path is use by firebase hosting to initialize everything, but we need to tell react to proxy this path to our local host when developing so we still get the hot loader

PR #520, opened 2020-05-13
refractor our mock data generation, add resources to mock data

mat10tng:


scheme/data.py
```diff
@@ -60,231 +59,264 @@ def genId():
 ]
 
 
+def any_time():
+    return int(time.time())
```
qdozaq, line 63: Looks like we need to multiply this by 1000.

other wise getting back `1589377870` which gives Fri Jan 02 1970  when run though `Date`

mat10tng: Date(1589377870)
"Wed May 13 2020 16:17:10 GMT+0200 (Central European Summer Time)"

scheme/data.json
```diff
@@ -1,3385 +1,1864 @@
 {
   "organizations": {
-    "1": {
+    "f112d652ecf13dacd9c78c11e1e7f987": {
```
qdozaq, line 3: We probably want this to stay "1" for now right?

mat10tng: yeah,

PR #538, opened 2020-05-14
fix bug where no auth crash home

mat10tng:


src/app/page/Home/Home.js
```diff
@@ -382,7 +382,11 @@ const SignInHeaderComponent = ({ history }) => {
  */
 const HomePage = ({ auth, history, profile }) => {
   const org = useOrganization();
+
   useFirestoreConnect(() => {
+    if (!auth.uid) {
+      return [];
+    }
```
jwu910, line 389: Don't connect if no auth object is provided?

mat10tng: yes

mat10tng: I mean, its a hook, so we cant have an if outside of it

jwu910: :+1:

PR #559, opened 2020-05-17
fix schema discrepancies

qdozaq:


src/app/model/schema.ts
```diff
@@ -30,7 +30,7 @@ export class OrganizationInterface {
   /* Firebase Id, created automatically*/
   uid!: string;
   /*Name of the organization - eg Feed Folks */
-  name!: string;
+  displayName!: string;
   /*Name of the organization chapter eg Feed Folks - Studio City
     (make this the same as 'name' if org doesn't have chapters) */
```
utunga, line 35: this comment should be fixed I guess.. ps not sure why we decided displayName over name but I guess as long as we are consistent.. 

actually on further thought I think name was gonna be eg "Feed Folks Coop LLC" or something and then displayName just "FeedFolks"... but this is fine as long as it works and if u think it's more consistent

src/app/page/Donate/DonationPage.jsx
```diff
@@ -147,9 +148,9 @@ function SuccessState({ classes, state }) {
           <Divider className={classes.marginVertical} />
           <Typography align="left" variant="subtitle2" gutterBottom>
             <em>
-              {org.name} is a registered 501(c)3 non-profit organization #{org.EINNumber}. Your
-              donation is tax deductible to the extent allowable by law. No goods or services were
-              provided by {org.name} in return for this contribution.
+              {org.displayName} is a registered 501(c)3 non-profit organization #{org.EINNumber}.
```
utunga, line 151: fwiw this has to get commented out for now anyways.. waiting to hear back formally but pretty sure it will be
> tax deductible yes 
> registered 501(c)3  no


 at least they will be looking for a fiscal sponsorship..

maybe this whole thing could be in an 'if' block based on existence of EIN number.. 

that said you don't have to address this comment in this PR

src/app/page/Donate/DonationPage.jsx
```diff
@@ -170,7 +171,8 @@ function ErrorState({ setState }) {
       </Typography>
       <Typography align="left" variant="body1" gutterBottom>
         Otherwise, our volunteers can help you donate through a different method, such as by check.
-        For more information please call: <a href={`tel:${org.phoneNumber}`}>{org.phoneNumber}</a>
+        For more information please call:{" "}
```
utunga, line 174: what does {" "} do ?

src/app/page/Donate/DonationPage.jsx
qdozaq: This should only show when the org actually has an EIN number set

src/app/model/schema.ts
qdozaq: yeh I like name better, but displayName is what we're using on all the other objects so just keeping consistent

src/app/page/Donate/DonationPage.jsx
qdozaq: eslint forces the jsx to the next line which removes the space that would naturally be there. this just manually adds that space back in

PR #568, opened 2020-05-18
Fix text overlapping for grid elements in missions

### LINKED ISSUE #507: UI bug on volunteer dashboard mobile
Seeing this when checking the site on an iphone both safari and chrome apps.

To test go here and sign in on mobile device: https://mutualaid-757f6.web.app/

appears to be an issue with grid/flex

IMG_4150

DrewThomasCorps:
fixes #507 

Screen Shot 2020-05-17 at 21 17 04

The grid-xs-true class's `flex-basis: 0` was causing the overlap in safari. I overrode this and set it to the default value of auto. I tested this in chrome and firefox and it looked good in both.


utunga:
👍

PR #569, opened 2020-05-18
Add build step to dev env setup

dyanawu:
I needed to run `npm run build` before I was able to start the dev server; this may or may not be necessary for everyone but I don't think it hurts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,7 @@ Here are a few geting started steps to help get set up for local development.
 7. Start your local development server with
 
    ```sh
+   npm run build
    npm run dev
    ```
 _NOTE_: On local development environment, you are always an Organizer.


qdozaq:
@dyanawu You shouldn't need this. The React app should be using the dev server here on localhost:3000 and proxying all the calls to firebase.

This is however what you would do if you wanted to test the app in its fully compiled state locally.
Running a build and then `firebase serve` or `npm run dev:firebase` will serve this static build which can viewed at localhost:5000

was there an issue with you not being able to access the app when it was not built first?

dyanawu:
@qdozaq Yeah - the first time I did run `npm run dev` it didn't run and suggested running `npm run build`, after which `npm run dev` did work. I don't know if it was potentially unrelated, but...

PR #573, opened 2020-05-18
routing fix

### LINKED ISSUE #571: Bugs in routing during foodbox request - success goes to a 404
**Steps to reproduce**

1. Go to  https://mutualaid-757f6.web.app/ and be signed out
1. Click on need help then food box
1. Choose some items and continue
1. Choose curbside pickup and continue
1. Enter name and phone number and then signup
_Validate if needed - I tried this with either validating phone number or with 'i cannot receive sms' checked, same result_
1. Click 'Find me a donation'
1. Click 'confirm'.

**Expected result**

The success page with the 'we're going to find you a donation' message.

**Actual result**

You get a 404 error page

image

--

Marking this as a critical bug because it may also disable the ability to pay by credit card but either way, this is a problem and we were hoping to go live with this aspect of the site Monday evening.

--

Also note... very possibly related.

If, at step 5 you do *not* check 'i cannot receive sms' then, after validation it takes you back to step 3 instead of proceeding on to step 6 as it used to do. (Once logged in you can now proceed through).

qdozaq:
fixes #571


src/app/page/Request/foodbox/ConfirmStep.jsx
```diff
@@ -88,8 +89,10 @@ function ConfirmStep({ dispatch, state }) {
     }
     try {
       const createdMission = await Mission.create(mission);
-      const redirect = isDonationRequest ? "donation" : "payment";
-      history.push(`/request/foodbox/success/${redirect}`);
+      const redirect = isDonationRequest
+        ? routes.request.success.donation
+        : routes.request.success.payment;
+      history.push(redirect);
```
jwu910, line 95: Is there a benefit to history.push vs using react routers `Redirect` ?

qdozaq: they're essentially the same `Redirect` just offers the "react-y" way to do it, but it still calls the history object behind the scenes. Though I suppose this should be `history.replace` to replicate exactly

amylo:
Looks good. I'll investigate slugs pathing later today. So we don't need the extra overhead.

PR #574, opened 2020-05-18
Feature/489 hamburger menu

amylo:


utunga:
perhaps a dumb question but has the actual list options for volunteers changed ? because like I'm really not sure that we want them all to be there right now some of them are a bit weird frankly and I don't think were ever actually designed.. ?

amylo:
> perhaps a dumb question but has the actual list options for volunteers changed ? because like I'm really not sure that we want them all to be there right now some of them are a bit weird frankly and I don't think were ever actually designed.. ?

They won't be shown all of those menu items at the simultaneously. This is how it looks for each type of user:

**Not logged In**
menu_not-logged-in

**Recipient**
menu_recipient

**Volunteer**
menu_volunteer

**Organizer**
menu_organizer

amylo:
> Checked it out locally and reviewed code all looks good.
> 
> I'm a tiny bit unsure of if we want 'i need help' to be visible when logged in. I mean we _want_ it to be there but I'm not entirely sure if we'll end up with a weird bug in that case...
> 
> I also added this bug but it was pre-existing so this is all good. Much better in fact.

We can remove it later if it is confusing for users.

PR #592, opened 2020-05-19
Convert and optimize images for web

jwu910:
Fixes #591<br/>Change webp to jpg for images to keep safari compatability in mind. Ran images through tinyjpg for lossless conversion for web.

PR #595, opened 2020-05-20
Bring PR #592 and PR #587 into production ?

utunga:
Just playing around with process here.. but wondering if it works to ask for confirmation that these features - which have already hit master - are good to go into the the release?

I mean pretty much right? Definitely need those webp fixes in there .. and I think the other changes also.. yeah?


jwu910:
I think the talk about the different branches that belong to studio city and feedfolks was lost in a 120+ message convo that happend at 3AM my time. lol

But yeah i think you can maybe just cherry pick the squashed commits to those respective branches, test and redeploy, should be good

I'm not 100% sure what we're comparing on this PR in regards to the state master is in, vs where the branch `release` was

utunga:
Hey @jwu910  sorry for the confusion. The difference between master and release  (which is also the same as the difference between master (now) and the last release to studio city).. well that difference is exactly what this PR is measuring. 

So, no need to squash or cherry pick - because I think these two prs (aka the HEAD of master) are good to go. 

I think I'm just gonna merge this and release to studiocity...

PR #597, opened 2020-05-20
Tag v002_mvp-1 release

utunga:
Bring PR #592 and PR #587 into production


utunga:
Dang adding a tag to release branch  triggers this PR back to master. SOOO CONFUSING . But is it best to do this so as to avoid history getting tangled up? I honestly dont know

PR #598, opened 2020-05-20
Fix volunteer profile signup not submitting

### LINKED ISSUE #472: [BUG] Sign up form submit buttons is missing functionality from sign up form and blocks volunteer flow
Steps to reproduce:
1. Go to Heroku development app[](https://resilience-app.herokuapp.com/)
2. Click on volunteer call to action button
3. Sign up with phone numer
4. Skip connect Faebook and Google accounts
5. Fill out sign up form and click submit
6. Observe the result

Expected result: 
Volunteer will enter their dashboard and see the available missions to volunteer for

Actual result:
The form submit button seems to be missing functionality and the form cannot be submitted.

Additional information: refreshing the page overrides this issue and options in side navigation for volunteers appear as normal.

DrewThomasCorps:
Fixes #472 

To test you can comment out line 45 of RoutePermissions.ts and use the 2223334444 phone number. (This route permissions blocks existing users from signing up again) or just use your own that you haven't used before. If you run `npm run start` you can see the values change in firebase.

I changed the organizationUid to a string to match the uid type in firebase.

In ConnectSocialMedia, PhoneAuth, and SignupSuccess, the prop types no longer matched what was being used, so I updated them (fixed some warnings in my IDE).

The loading state was not being used in SignupScene so I removed it.

I updated the handleSuccess functions in SignupScene to pass the data to updateUser. This prevents the user from updating before the value state finished updating. After the user is successfully updated, then the state is updated along with the page.

I passed the setActiveState in as a callback to updateUser so it could update it after finishing updating the user.

`updateUser()` didn't previously update the user. Now it does.

Lastly I moved the logic for getting the address into a separate function to improve readability. (2 sets of try catches seemed extra).

### DIFF
--- a/src/app/model/User.ts
+++ b/src/app/model/User.ts
@@ -18,7 +18,7 @@ const defaultUserData: UserInterface = {
   phoneNumber: "",
   email: "",
   location: defaultLocation,
-  organizationUid: 0,
+  organizationUid: "",
   isVolunteer: false,
   isOrganizer: false,
   volunteerDetails: {
--- a/src/app/page/Signup/SignupScene.jsx
+++ b/src/app/page/Signup/SignupScene.jsx
@@ -1,5 +1,5 @@
 import PropTypes from "prop-types";
-import React, { useState } from "react";
+import React, { useContext, useState } from "react";
 
 import Snackbar from "../../component/Snackbars/Snackbar";
 import useForm from "../../hooks/useForm";
@@ -13,6 +13,7 @@ import SignupSuccessPage from "./SignupSuccess";
 import UserProfilePage from "./UserProfile";
 import { withRouter } from "react-router-dom";
 import { routes } from "../../routing";
+import { SnackbarContext } from "../../component/Snackbars/Context";
 
 const Tabs = {
   GET_STARTED: "volunteer call to action to sign up",
@@ -28,7 +29,7 @@ const Tabs = {
 function SignupScene(props) {
   const { handleChange, setValues, values } = useForm(User.defaultData);
   const [activeTab, setActiveTab] = useState(Tabs.GET_STARTED);
-  const [errorSnackbarMessage, setErrorSnackbarMessage] = useState(false);
+  const snackbarContext = useContext(SnackbarContext);
   const org = useOrganization();
 
   function getPayload() {
@@ -55,7 +56,6 @@ function SignupScene(props) {
       uid: user.uid,
     };
     await updateUser(data, () => setActiveTab(Tabs.CONNECT));
-    return false;
   }
 
   async function handleConnectSuccess(user) {
@@ -67,39 +67,30 @@ function SignupScene(props) {
       uid: user.uid,
     };
     await updateUser(data, () => setActiveTab(Tabs.PROFILE));
-    return false;
   }
 
-  async function handleProfileSuccess() {
-    await updateUser({ ...getPayload() }, () => setActiveTab(Tabs.SUCCESS));
+  async function handleProfileUpdate() {
+    await updateUser({ ...getPayload() }, () => {
+      setActiveTab(Tabs.SUCCESS);
+      snackbarContext.show({
+        message: "Congratulations, your volunteer account has been created!",
+        type: "success",
+      });
+    });
   }
 
   async function updateUser(payload, callback) {
-    payload = await getPayloadWithAddress(payload);
     try {
-      User.update(payload.uid, payload).then(() => {
-        setValues({ ...payload });
-        callback();
-      });
-    } catch (error) {
-      setErrorSnackbarMessage(error);
+      await User.update(payload.uid, payload);
       setValues({ ...payload });
       callback();
-    }
-  }
-
-  async function getPayloadWithAddress(payload) {
-    let location = payload?.location;
-    try {
-      if (location.address) {
-        const data = await addressLookup(location.address);
-        location = { ...location, lat: data.lat, lng: data.long };
-        payload = { ...payload, location: location };
-      }
     } catch (error) {
-      setErrorSnackbarMessage(error);
+      snackbarContext.show({
+        message: `Error while submitting request. Please try again: ${error.message}`,
+        type: "error",
+      });
+      setValues({ ...payload });
     }
-    return payload;
   }
 
   let Active;
@@ -131,7 +122,7 @@ function SignupScene(props) {
       Active = (
         <UserProfilePage
           handleChange={handleChange}
-          onSubmit={handleProfileSuccess}
+          onSubmit={handleProfileUpdate}
           values={values}
         />
       );
@@ -142,24 +133,7 @@ function SignupScene(props) {
       break;
   }
 
-  const showSuccessSnackbar = !errorSnackbarMessage && activeTab === 4;
-  const snackbarOpen = errorSnackbarMessage || showSuccessSnackbar;
-  const snackbarType = showSuccessSnackbar ? "success" : "error";
-  const snackbarMessage = showSuccessSnackbar
-    ? "Your account request has been successfully submitted and is pending approval"
-    : `Error while submitting request. Please try again. ${errorSnackbarMessage}`;
-  return (
-    <>
-      {Active}
-      <Snackbar
-        open={snackbarOpen}
-        handleClose={() => setErrorSnackbarMessage(false)}
-        type={snackbarType}
-        message={snackbarMessage}
-        autoHideDuration={4000}
-      />
-    </>
-  );
+  return <>{Active}</>;
 }
 
 SignupScene.propTypes = {
--- a/src/app/page/Signup/SignupScene.jsx
+++ b/src/app/page/Signup/SignupScene.jsx
@@ -29,7 +29,6 @@ function SignupScene(props) {
   const { handleChange, setValues, values } = useForm(User.defaultData);
   const [activeTab, setActiveTab] = useState(Tabs.GET_STARTED);
   const [errorSnackbarMessage, setErrorSnackbarMessage] = useState(false);
-  const [loading, setLoading] = useState(false);
   const org = useOrganization();
 
   function getPayload() {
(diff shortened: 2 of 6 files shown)


src/app/model/User.ts
```diff
@@ -18,7 +18,7 @@ const defaultUserData: UserInterface = {
   phoneNumber: "",
   email: "",
   location: defaultLocation,
-  organizationUid: 0,
+  organizationUid: "",
```
jwu910, line 21: This is the 2nd or 3rd time I've seen someone change this between a string and a number. Have we been able to identify what this should be, and curious if we've checked how this propogates to the rest of the app.

DrewThomasCorps: When I look at the firebase console, the organizationUid is a string. And the organizations' uid is a string: "1". I think I was getting an error which is why I had to change it, but I can't remember for sure.

DrewThomasCorps: I can confirm that an error popped up when I changed this back to a number on my local copy.

mat10tng: This should be a string, because it is going to refer to a database fk, so that is not ever going to be a number....

src/app/page/Signup/SignupScene.jsx
```diff
@@ -46,59 +45,65 @@ function SignupScene(props) {
...
     let location = payload?.location;
     try {
       if (location.address) {
         const data = await addressLookup(location.address);
         location = { ...location, lat: data.lat, lng: data.long };
+        payload = { ...payload, location: location };
       }
     } catch (error) {
-      console.log("ERROR WHEN GETTiNG LOCATION", error);
+      console.log("ERROR WHEN GETTING LOCATION", error);
```
mat10tng, line 100: console.error, if we want to use log

src/app/page/Signup/SignupScene.jsx
```diff
@@ -46,59 +45,65 @@ function SignupScene(props) {
...
       console.log("address: ", location.address);
```
mat10tng, line 101: remove this

DrewThomasCorps: I removed it and set the error snackbar message. We don't expect it to occur, but this way we'll see something if it does. (I also added an await in the updateUser)

src/app/page/Signup/SignupScene.jsx
jwu910: Can we use a Snackbar here instead?

DrewThomasCorps: It does, check the latest commit

src/app/page/Signup/SignupScene.jsx
```diff
@@ -97,8 +97,7 @@ function SignupScene(props) {
         payload = { ...payload, location: location };
       }
     } catch (error) {
-      console.log("ERROR WHEN GETTING LOCATION", error);
-      console.log("address: ", location.address);
+      setErrorSnackbarMessage(error);
```
jwu910, line 100: Would it be okay if you have another commit that removes this old implementation of snackbar, and use the new snackbar using the callback provided in the snackbar context? cleaner/newer pattern and this should be throwing deprecation warnings in the console

src/app/page/Signup/SignupScene.jsx
```diff
@@ -75,7 +75,7 @@ function SignupScene(props) {
   }
 
   async function updateUser(payload, callback) {
-    payload = getPayloadWithAddress(payload);
+    payload = await getPayloadWithAddress(payload);
     try {
       User.update(payload.uid, payload).then(() => {
```
jwu910: you'll want all of your awaits inside your try catch, otherwise if getPayloadWithAddress fails, it won't get caught by this error handling logic but by the next (if any) error mgmt logic up the tree

src/app/model/User.ts
jwu910: Cool thanks for clarifying

src/app/page/Signup/SignupScene.jsx
DrewThomasCorps: Definitely! I’ll do that tonight. I’ll also do that to the other snack bar in the component when updating the user.

src/app/page/Signup/SignupScene.jsx
```diff
@@ -46,59 +45,64 @@ function SignupScene(props) {
...
+    payload = await getPayloadWithAddress(payload);
+    try {
+      User.update(payload.uid, payload).then(() => {
+        setValues({ ...payload });
+        callback();
+      });
+    } catch (error) {
+      setErrorSnackbarMessage(error);
+      setValues({ ...payload });
+      callback();
```
jwu910, line 87: Should this callback get called regardless if this is a pass or fail?

src/app/page/Signup/SignupScene.jsx
```diff
@@ -46,59 +45,64 @@ function SignupScene(props) {
...
-    const payload = getPayload();
 
+  async function handleProfileSuccess() {
+    await updateUser({ ...getPayload() }, () => setActiveTab(Tabs.SUCCESS));
+  }
+
+  async function updateUser(payload, callback) {
+    payload = await getPayloadWithAddress(payload);
+    try {
+      User.update(payload.uid, payload).then(() => {
```
jwu910, line 80: I think we should stick with one pattern regarding async/await or promise chaining.

Couldnt this be: 

```
await User.update(payload.uid, payload)
```
?

src/app/page/Signup/SignupScene.jsx
```diff
@@ -46,59 +45,64 @@ function SignupScene(props) {
...
+    };
+    await updateUser(data, () => setActiveTab(Tabs.PROFILE));
     return false;
   }
-  async function updateUser() {
-    const payload = getPayload();
 
+  async function handleProfileSuccess() {
+    await updateUser({ ...getPayload() }, () => setActiveTab(Tabs.SUCCESS));
+  }
```
jwu910, line 75: So my comments might be out of order, was reading and trying to follow the logic and renaming changes.

I think this would make more sense to be called `handleUpdateSubmit` or something. because the responsibility of this is to invoke the `updateUser` method, and not actually do on a `ProfileSuccess`

Also might be foggy on the async await, maybe @qdozaq or @mat10tng can confirm, but you wrote updateUser as an async function that's actually invoking the call to the API with User.update. I'm not sure we need async declared for this function.

jwu910 (changes requested):

src/app/page/Signup/SignupScene.jsx
```diff
@@ -46,59 +45,64 @@ function SignupScene(props) {
...
+      ...getPayload(),
       displayName: user.displayName,
       email: user.email,
       photoURL: user.photoURL,
-    });
-    updateUser();
-    setActiveTab(Tabs.PROFILE);
+      uid: user.uid,
+    };
+    await updateUser(data, () => setActiveTab(Tabs.PROFILE));
```
qdozaq, line 69: Yeh so for these we could just call `User.update` directly.

two ways to handle
async/await:
```js
await User.update(data.uid, data);
setActiveTab(...)
```
promises:
```js
User.update(data.uid, data).then(() => setActiveTab(...));
```

I actually prefer the promise the version here as then you don't have to make this handler an async call.
Also not sure why these return `false` but we can leave handlers as void.

qdozaq: Oh I see `updateUser` is supposed to be a general handler for all of these.. I'll comment over there

jwu910: I've always felt .then chaining was a little easier to read until you get into callback hell and its like .then().then().then().then().then()... but _then_ you have a different problem.

src/app/page/Signup/SignupScene.jsx
qdozaq: yeh lets use await here. and wrap all the logic with the try/catch

src/app/page/Signup/SignupScene.jsx
qdozaq: I would say, probably not

src/app/page/Signup/SignupScene.jsx
DrewThomasCorps: Thanks, will do!

src/app/page/Signup/SignupScene.jsx
qdozaq: Lol pretty much, as soon as you have multiple promises from different sources you should just use await*

*unless they can be done parallel lol

DrewThomasCorps: I wasn't sure on the returning false either. I didn't know if it had something to do with the firebase callback so I left it. I can't find any documentation for `firebaseui` in the FirebaseAuthUI component.

src/app/page/Signup/SignupScene.jsx
DrewThomasCorps: I'm about to push my changes. To use await you need async declared, so this is the one thing I didn't change. If you want though I could switch it to .then().

src/app/page/Signup/SignupScene.jsx
```diff
@@ -46,59 +46,54 @@ function SignupScene(props) {
...
-  async function updateUser() {
-    const payload = getPayload();
 
-    let location = payload?.location;
+  async function updateUser(payload, callback) {
     try {
-      if (location.address) {
-        const data = await addressLookup(location.address);
-        location = { ...location, lat: data.lat, lng: data.long };
-      }
```
DrewThomasCorps, line 80: Just so it's clear, this could be removed because the component used in the profile section fills this data in (and `addressLookup` wasn't working anyway)

utunga:
Just an FYI I get this when I run locally - localhost -  I asume this is because the firebase functions dont run locally.

How does one work around this problem on localhost? Use the test accounts perhaps?

image
image
image

Incidentally I see that firebase functions will start to be only available to accounts which charge money in future. Is there any way we can do this that doesnt rely on firebase functions?

utunga (approved):
Requested changes have been addressed. Drew can resolve conversations/issues that remain but up to him to do that. Approved so its not a blocker

src/app/page/Signup/SignupScene.jsx
utunga: very happy to stick with await fwiw.. 

await vs .then() is a style question honestly

src/app/page/Signup/SignupScene.jsx
DrewThomasCorps: I changed it to await, and removed the getPayloadWithAddress so it's all in the try/catch.

src/app/page/Signup/SignupScene.jsx
DrewThomasCorps: It's left as a wrapper but I switched everything to await.

utunga (approved):

DrewThomasCorps:
@jwu910 I believe this has been approved, but your change request is blocking it. Would you mind re-reviewing? I'd be happy to make more changes if needed.

src/app/page/Signup/SignupScene.jsx
```diff
@@ -46,59 +46,54 @@ function SignupScene(props) {
...
-        setActiveTab(Tabs.SUCCESS);
+      snackbarContext.show({
+        message: `Error while submitting request. Please try again: ${error.message}`,
+        type: "error",
       });
-    } catch (error) {
-      setLoading(false);
-      setErrorSnackbarMessage(error);
-      setActiveTab(Tabs.SUCCESS);
+      setValues({ ...payload });
```
jwu910, line 92: Do we want these values set if there is an error ?

src/app/page/Signup/SignupScene.jsx
```diff
@@ -125,33 +120,20 @@ function SignupScene(props) {
...
-      <Snackbar
-        open={snackbarOpen}
-        handleClose={() => setErrorSnackbarMessage(false)}
-        type={snackbarType}
-        message={snackbarMessage}
-        autoHideDuration={4000}
-      />
-    </>
-  );
+  return <>{Active}</>;
```
jwu910, line 136: We won't need the fragment here since there is only one element being returned.

If i recall correctly, since Active is assigned to a component,  you can return `<Active />` or just `Active`

src/app/page/Signup/SignupScene.jsx
```diff
@@ -46,59 +46,54 @@ function SignupScene(props) {
...
+    await updateUser(data, () => setActiveTab(Tabs.PROFILE));
+  }
+
+  async function handleProfileUpdate() {
+    await updateUser({ ...getPayload() }, () => {
+      setActiveTab(Tabs.SUCCESS);
+      snackbarContext.show({
+        message: "Congratulations, your volunteer account has been created!",
+        type: "success",
+      });
```
jwu910, line 78: I feel that some of the nested await function calls have gotten a little confusing.

I think we can consolidate some of these handlers to remove some of the jumping around of the logic to something like this (not exact)

```
// Update profile function gets invoked
async function handleProfileUpdate(payload) {
  try { // Everything should only happen upon a successful user.update
    await User.update(payload.uid, payload); // This promise should be resolved before this logic can continue. If there is an error thrown here, caught by catch()
    setValues(payload); // setValues after user is updated
	setActiveTab(Tabs.SUCCESS);
    snacbarContext.show({ // snackbar show success after everything above has run
      message: "Success", type: "success"
    })
  } catch (error) { // show error if there is an error updating user.
    snacbarContext.show({
      snackbarContext.show({message:"Error", type:"error"})
    })
  }
}

```

I think something like this would make more sense rather than passing around and invoking the `callback` param that we're passing into `updateUser`.

jwu910 (changes requested):
The comment about the organization around the async/await calls are my only concern. If it works and logic makes sense to everyone else, then I'll just remove my review

src/app/page/Signup/SignupScene.jsx
DrewThomasCorps: These all make sense, I'll work on it tonight!

DrewThomasCorps:
Just an update: I'm running into a bug I thought I had fixed where it refreshes the PhoneAuth page rather that calls the callback. There might be a race condition somewhere. I'll work on it more tomorrow and ask for help if I can't figure it out.

PR #600, opened 2020-05-20
mission delivered and mission completed view

mat10tng:


src/app/layout/NavigationDrawer/NavigationDrawer.js
```diff
@@ -22,10 +20,11 @@ import { useStyles } from "./NavigationDrawer.style";
 import { AppLink, routes } from "../../routing";
 import { PERMISSIONS, UserPermissionsService } from "../../model/permissions";
 
+//Note: ListItemIcon always require a child
 const MenuItem = ({ classes, icon, text, to }) => (
   <AppLink to={to} className={classes.link}>
     <ListItem button>
-      <ListItemIcon>{icon}</ListItemIcon>
+      <ListItemIcon>{icon ? icon : <div />}</ListItemIcon>
```
jwu910, line 27: If this always requires a child, can we set a fallback icon then or just make it a required prop of MenuItem?

mat10tng: it is because of the design.

jwu910: OH yeah, in this case then, i would just do
```
{icon && (
  <ListItemIcon>{icon}</ListItemIcon>
)}
```
To avoid having to create an empty div

mat10tng: we need the padding, unfortunately, otherwise the text would align to left side

karenkun: can we not use some css to do it ??  (put a classname for the menulistitem that will put a padding there for when you want "no icon" but still keep the visual spacing) ?

mat10tng: yes we can do that as well, but -> it have to have the exact same css as the ListIconItem div any how...

jwu910: CSS changes would be preferrable to additional DOM elements i think

mat10tng: alright, fair enough. I am convinced. On it.

mat10tng: done

PR #601, opened 2020-05-20
 Attempt to force index.html to always check for more recent version

### LINKED ISSUE #419: What the heck is up with the cache?
.. if we release an upgrade.. I will sometimes see old code for days afterward.

Is there a webpack setting or http-cache directive or combination that can force stale code to expire while still caching until it changes?

This is sometimes called 'cache-busting' and I coulda sworn its something that came out of the box with webpack so not sure why it doesn't work?

But I definitely have seen old pages - at least two or three days old - and then hit shift refresh and got the latest.

utunga:
Attempt to fix #419 

For a brief moment of time this explanation of cache headings made sense
https://csswizardry.com/2019/03/cache-control-for-civilians/

The intention here is that **only** the root page / (aka index.html)
has the cache-control:no-cache header set.. and that should apparently
result in browsers always checking for that one file (though
if its up to date they can still use cache even though, well anyway
read the article)

As to whether this actually works.. we shall find out!

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -33,7 +33,7 @@
       "source": "/",
       "headers": [ {
         "key": "Cache-Control",
-        "value": "max-age=300"
+        "value": "no-cache"
       } ]
     } ]
   },


jwu910:
Just started reviewing :)

jwu910:
Is this possible to test? after pulling the code down, realized unless i spin up my own server serving production build files, i can't really reproduce the same environment this is concerned with.

utunga:
I mean no-cache sounds pretty cool to me And it does sort of say that it's the recommended approach for a dynamic website... bottom line is I think we have to deploy and test to really know for sure. (CDNs and hosting and production builds etc all being part of the mix) .

During rapid dev no-store might be more appropriate.. though.

I don't really know. Main thing is one or the other has to be better than what we have now let's get one up and try it.

Happy to switch to no-store if u wanna start w that.

utunga:
uhm really keen to get this change in - its not harmful can we please test on staging?

utunga:
I sneakily deployed this to both staging and production on top of the [v003_mvp-3](https://github.com/factn/resilience-app/tree/v003_mvp-1) tag.  This is because I live in terror trying to fix a bug and then not being able to force a cache refresh, and know that the sooner we fix a caching problem the less likely we'll have hair-pulling frustration down the track.

utunga:
FYI @jwu910  I sneakily deployed this to both staging and production on top of the [v003_mvp-1](https://github.com/factn/resilience-app/tree/v003_mvp-1) tag.

I know that is terrible breaking of 'the rules' but sometimes it is right to break the rules. I am living in terror that at some point we'll have a bug on live and not be able to fix it due to caching. So the sooner we figure out this caching thing the more likely we'll be able to avoid hair pulling frustrations down the track. 

So.. once the current code has finally expired out of caches (or been forced out) then it will be possible to deploy new code to staging and test if it actually updates rapidly enough with a `no-cache` header set.

PR #602, opened 2020-05-21
Add og meta tags for social sharing

### LINKED ISSUE #596: Fix social sharing - currently it says 'website created using create-react-app'
This kinda sucks. **Currently** when you share a link to our site you see..

In facebook:
image

In slack:
image

Social integration is a whole thing, obviously, but as a first step it t would be good if we could just fix our og:graph meta tags.  https://ogp.me/

We don't have to go crazy but it might be a start to use the organization data for this...
org.title, org.tagline and org.logo would be a good start for our OG:meta content 

Also see this.. https://developers.facebook.com/tools/debug/

Maybe whack 'powered by resilience' in there somewhere? Or maybe not

DrewThomasCorps:
resolves #596 
Screen Shot 2020-05-20 at 20 43 05
This adds og tags using the organization details. (the local-org.firebaseapp.com won't show up when this is live, I just used that to test the tags on localhost)

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -9,20 +9,20 @@
 
     <!-- Twitter Card data -->
     <meta name="twitter:card" content="summary">
-    <meta name="twitter:title" content="Feed Folks">
+    <meta name="twitter:title" content="Studio City - Feed Folks">
     <meta name="twitter:description" content="We're a grassroots team in Studio City, CA getting fresh farm produce to our neighbors in need.">
     <meta name="twitter:image:src" content="https://firebasestorage.googleapis.com/v0/b/mutualaid-757f6.appspot.com/o/images%2Ffeedfolks__logo.png?alt=media&token=6b1e803d-9b19-4847-a849-4b4dbdde2395">
 
     <!-- Open Graph data -->
-    <meta property="og:title" content="Feed Folks" />
+    <meta property="og:title" content="Studio City - Feed Folks" />
     <meta property="og:type" content="website" />
     <meta property="og:url" content="https://mutualaid-757f6.web.app" />
     <meta property="og:image" content="https://firebasestorage.googleapis.com/v0/b/mutualaid-757f6.appspot.com/o/images%2Ffeedfolks__logo.png?alt=media&token=6b1e803d-9b19-4847-a849-4b4dbdde2395" />
     <meta property="og:description" content="We're a grassroots team in Studio City, CA getting fresh farm produce to our neighbors in need." />
-    <meta property="og:site_name" content="Feed Folks" />
+    <meta property="og:site_name" content="Studio City - Feed Folks" />
 
     <!-- Schema.org markup for Google+ -->
-    <meta itemprop="name" content="Feed Folks">
+    <meta itemprop="name" content="Studio City - Feed Folks">
     <meta itemprop="description" content="We're a grassroots team in Studio City, CA getting fresh farm produce to our neighbors in need.">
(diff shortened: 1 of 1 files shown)


jwu910:
Have we checked on other popular Social media platforms for sharing previews?

E.g.: Twitter, Facebook?

DrewThomasCorps:
I did not. I don't have either. It appears that if `twitter:` tags are missing it falls back to the `og:` tags. https://developer.twitter.com/en/docs/tweets/optimize-with-cards/overview/markup
I don't currently see a need for any of the other twitter tags, but can add them if wanted.

DrewThomasCorps:
Screen Shot 2020-05-20 at 21 35 00
Screen Shot 2020-05-20 at 21 34 25
Facebook and twitter screenshots.

utunga:
What you've done is great and I'd be super tempted to merge it as is - but is there any chance you can go a step further and  - in the header - call the 'useOrganization()' hook to get access to the 'org' object and then pull this data direct from the organisation? 

Some of this may not be 'in' the org object yet but where it is best to pull it from there (and where it isn't we should add it - but i guess that can be out of scope for this PR). 

Here's the values we deployed for studiocity... you should be able to access all this as properties of 'org' object..


```

"uid": "1",
      "displayName": "Feed Folks",
      "chapterName": "Feed Folks - Studio City",
      "tagline": "Neighbors Helping Neighbors",
      "quickInfo": "We're a grassroots team in Studio City, CA getting fresh farm produce to our neighbors in need.",
      "logoURL": "https://firebasestorage.googleapis.com/v0/b/resilienceapp-6a0dd.appspot.com/o/images%2Ffeedfolks_logo.png?alt=media&token=e9278f38-5228-403e-815d-8b5742d4afa0",
      "contactEmail": "help@studiocitync.org",
      "contactPhoneNumber": "",
      "location": {
        "address": "Studio City, CA",
        "label": "",
        "lat": 34.1483989,
        "lng": -118.3961877
      },
```

utunga:
as i say pretty tempted to just merge as is if that seems like a lot of work. this whole card is a bit of sneaky scope creep on my aprt relative to our original MVP0 goal to be honest with myself.

DrewThomasCorps:
@utunga I was going to do that, but since it's in the index.html page it's outside of the React logic. If we set up some form of server-side rendering I could replace placeholder values before serving it or use a library like react helmet. https://github.com/nfl/react-helmet

qdozaq:
Yeh @miles it's a little out of scope. getting dynamic data to populate for search engines is going to take a little more work.

I think this is fine for now. Though we may want to switch to something more generic once we have more than one org.

For those curious about what would need to be done to get this to work here's a nice article that sums it up: https://medium.com/@jalalio/dynamic-og-tags-in-your-statically-firebase-hosted-polymer-app-476f18428b8b

utunga:
Yup content is good

utunga:
Actually, @DrewThomasCorps  can you change the title to "Studio City - Feed Folks"  instead of just "Feed Folks"?

Right now they honestly dont know what the heck they are doing and might hcange their mind on content at any time but at this point this looks good.

utunga:
or.. better switch it up so its all in the db .. and therefore easy to change.

qdozaq:
> or.. better switch it up so its all in the db .. and therefore easy to change.

@utunga Can't do that unless we implement the functions in the article I linked before, which we I think we should hold off on right now (unless someone really wants to tinker with this)

@DrewThomasCorps lets do the "Studio City - Feed Folks" fix and then we can merge it in. Thanks!

utunga:
What @qdozaq  said. tx

utunga:
Dang it my apologies for not reading the whole thread! I promise to be better in future.

DrewThomasCorps:
I agree that this should only be a temporary fix, but it's a step up from 'created using create-react-app'. I'm gonna play around with the link @jwu910 sent and see if I can get it implemented. It looks like it would be harder to test without deploying though.

PR #605, opened 2020-05-21
Recipient dashboard

qdozaq:
submitted
Annotation 2020-05-21 185605


src/app/page/Recipient/Dashboard.tsx
```diff
@@ -28,31 +28,41 @@ const useStyles = makeStyles((theme) => ({
...
+      let submitted: MissionInterface[] = [];
+      let completed: MissionInterface[] = [];
 
       missions.forEach((mission) => {
         completedStatus.includes(mission.status)
           ? completed.push(mission)
           : submitted.push(mission);
       });
 
+      submitted.sort(sortByDate);
```
mat10tng, line 63: you could sort the missions first, then both submitted and completed would be sort by default when they are pushed in

src/app/page/Recipient/Details.tsx
```diff
@@ -0,0 +1,135 @@
...
+    <>
+      <Divider />
+      <Box margin="1rem">
+        {mission.deliveryType === "delivery" ? (
+          <>
+            <DetailSection
+              header="Address"
+              content={
+                <Typography variant="body1" gutterBottom>
+                  {mission.deliveryLocation.address}
```
mat10tng, line 21: mission?.

qdozaq: I can, but typescript guarantees that mission will exist here

src/app/model/Mission.ts
```diff
@@ -265,7 +266,7 @@ class Mission extends BaseModel {
    * @param {string} missionUid - mission
    * @param {object} data- updated data
    */
-  update(missionUid: string, data: object) {
+  update(missionUid: string, data: Partial<MissionInterface>) {
```
qdozaq, line 269: Special TS wrapper to make all properties optional

src/app/routing/components/AppRoute.tsx
```diff
@@ -1,26 +1,28 @@
...
 const loadingStyles = makeStyles((theme) => ({
   root: {
     width: "100%",
   },
 }));
 
-function AppRoute({ children, component, ...rest }) {
+type Props = RouteProps;
+
+function AppRoute({ children, component, ...rest }: Props) {
```
qdozaq, line 19: @amylo I converted this component to typescript. lemme know if I'm missing anything

src/app/model/schema.ts
```diff
@@ -215,7 +216,7 @@ export interface MissionInterface {
   deliveryLocation: Location; // default to recipient location
   deliveryConfirmationImage: string;
   deliveryNotes: string;
-  deliveryType: string;
+  deliveryType: "delivery" | "curbside";
```
jwu910, line 219: :eyes:

src/app/page/Recipient/Details.tsx
```diff
@@ -0,0 +1,135 @@
...
+        )}
+      </Box>
+    </>
+  );
+}
+
+function DetailSection({ content, header }: { content: ReactNode; header: ReactNode }) {
+  return (
+    <Box marginBottom="1rem">
+      {React.isValidElement(header) ? header : <Typography variant="h5">{header}</Typography>}
```
jwu910, line 65: Oh interesting validation to check for jsx vs just a text string?

src/app/page/Recipient/Details.tsx
```diff
@@ -0,0 +1,135 @@
...
+    if (!isEditing) {
+      setIsEditing(true);
+      return;
+    }
+
+    Mission.update(mission.uid, { deliveryNotes })
+      .then(() => {
+        setIsEditing(false);
+      })
+      .catch((e) => {
```
jwu910, line 87: Do we want to provide `e.message` for any reason to the user or organizer? Is it possible Mission.update would have any valuable info that they could resolve?

src/app/model/schema.ts
qdozaq: its like a poor mans enum. Really nice cause it auto fills these options when either filling the prop or using it to compare

src/app/page/Recipient/Details.tsx
qdozaq: yah!

src/app/page/Recipient/RequestsList.tsx
```diff
@@ -0,0 +1,235 @@
...
+  const [open, setOpen] = useState(false);
+  const toggle = () => setOpen(!open);
+
+  return (
+    <Card className={classes.card}>
+      <TopBar
+        status={mission.status}
+        fundedStatus={mission.fundedStatus}
+        createdDate={mission.createdDate}
+      ></TopBar>
```
jwu910, line 109: Just syntax, but can we just self close if there are no children?

src/app/page/Recipient/RequestsList.tsx
```diff
@@ -0,0 +1,235 @@
...
+  ) : status === Status.delivered ? (
+    <NotificationImportant {...rest} />
+  ) : status === Status.succeeded ? (
+    <CheckCircle {...rest} />
+  ) : fundedStatus === FundedStatus.notfunded ? (
+    <HourglassFull {...rest} />
+  ) : (
+    <FiberNew {...rest} />
+  );
+}
```
jwu910, line 154: My eyes! haha jk

Can we just use a hash map(?) like this:

```
const ICON_STATUS = {
  [Status.started]: PlayCircleFilled,
  [Status.delivered]: <NotificationImportant,
  [Status.succeeded]: CheckCircle,
  [FundedStatus.notfunded]: HourglassFull
  ['default']: FiberNew
}

// and then use it something like
const StatusIcon = (status) => {
	if (!(Object.keys(ICON_STATUS).contains(status)) {
		return ICON_STATUS['default']
	}

    return ICON_STATUS[status]
}

//and then 
//...

<StatusIcon [...your props ]/>
```

Some of these ternary chains are really hard to follow

src/app/page/Recipient/Details.tsx
qdozaq: I wish we had a logging service... I can `console.error` it but I don't think customers need to see the raw message

src/app/page/Recipient/RequestsList.tsx
qdozaq: I wanted to do a map or even just a switch but I need to account for two different status objects which causes problems... like when would fundedstatus override the regular status?

Would and `if/else` look cleaner?

```ts
if (status === STATUS...) return <Icon/>
if (status === STATUS...) return <Icon/>
if (status === STATUS...) return <Icon/>
if (status === STATUS...) return <Icon/>
if (fundedstatus === FUNDEDSTATUS...) return <Icon/>
```

src/app/model/schema.ts
amylo: Might as well actually define String Enum lol.... Cuz then we can share and reuse elsewhere when needed. Don't want to give people reason to cooy-pasta.

src/app/page/Recipient/Details.tsx
amylo: We probably have bigger issues, if a required property is missing 😉 should trust TypeScript to enforce the requirement.

src/app/model/schema.ts
qdozaq: Haha okay. Fair enough

src/app/page/Recipient/RequestsList.tsx
mat10tng: I would prefer switch due to both:
1. Switch is made for this use case, map is just a convenient way.
2. You actually create an object every time you render using the map approach, we had this discussion earlier.

qdozaq: neither the switch nor map will work unless we can consolidate all statuses to single property.
Need to be able to test for both status and funded status which I can't really directly do in a switch.
Gonna try the `if` block approach

src/app/model/schema.ts
jwu910: I threw the eyes in because this was changed from `"delivery" | "curbside"` to `string` a few days ago and now back to this lol

qdozaq: Oh lol...
It should definitely be a limited set of values as we're using it to determine UI state.

src/app/model/Mission.ts
```diff
@@ -360,6 +362,7 @@ class Mission extends BaseModel {
       volunteerUid: userUid,
       volunteerDisplayName: user.displayName,
       volunteerPhoneNumber: user.phoneNumber,
+      volunteerPhotoUrl: user.photoURL,
```
mat10tng, line 365: volunteerPhotoURL is more consistent

qdozaq: done

src/app/page/Recipient/RequestsList.tsx
```diff
@@ -0,0 +1,227 @@
...
+
+function TopBarText({ createdDate, fundedStatus, status }: any) {
+  const date = new Date(createdDate).toLocaleDateString();
+
+  if (status === Status.started) return <>In progress | Accepted</>;
+  if (status === Status.delivered) return <>It's here! Confirm delivery below</>;
+  if (status === Status.succeeded) return <>Delivery Confirmed</>;
+  if (fundedStatus === FundedStatus.notfunded) return <>Awaiting donation | Submitted {date}</>;
+  return <>Submitted {date}</>;
```
jwu910, line 157: Lol Hmm i mean it... works.. It's actually much easier to read with this many if's. I just feel like this can be a littler DRY-er somehow... But i think we're good for now. 

Does this need to return text in a fragment though? Should that just be a string?

qdozaq: I was tempted to share the Icon and text state in the same component as they share the same return states, but that isn't necessarily guaranteed going forward as the designs don't really show all the possible combinations that the top bar can be in. Keeping it separate should make future changes easier if/when these states diverge.

I had initially used fragment as typescript has an issue since `string` isn't a jsx constructor. Apparently still an ongoing issue (https://github.com/DefinitelyTyped/DefinitelyTyped/issues/20544)

Went back and was able to cheat it with `any` though so guess that works lol

jwu910:
Ticket #?

PR #631, opened 2020-05-25
this is due to how npm package handler + ts

mat10tng:
image
https://www.typescriptlang.org/tsconfig#typeRoots

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -1,6 +1,9 @@
 {
   "functions": {
-    "predeploy": ["npm --prefix ./functions run build"],
+    "predeploy": [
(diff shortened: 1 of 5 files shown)


firebase.json
```diff
@@ -1,9 +1,6 @@
 {
   "functions": {
-    "predeploy": [
-      "npm --prefix \"$RESOURCE_DIR\" run lint",
```
qdozaq, line 4: Why remove the linting stage? Will type errors still be caught on build?

mat10tng: Added it back, it was during my debug checking process.

utunga:
PS was able to deploy just now yay

PR #634, opened 2020-05-26
allow volunteer to change pickup and delivery instructions

mat10tng:


src/app/component/MissionComponent/MissionDetailsCard/MissionDetailsCard.jsx
```diff
@@ -77,11 +141,17 @@ const MissionDetailsCard = ({ mission, photoDisabled }) => {
...
   const classes = useStyles();
 
-  if (mission.status === "started") {
+  if (mission.status === Mission.Status.started) {
     photoDisabled = false;
   } else {
     photoDisabled = true;
   }
+
+  let notesEditDisabled = false;
```
qdozaq, line 150: This should come from a `useState` call

qdozaq: err.. Oh I see why you did this.

Could it instead be:

```js
const notesEditDisabled = [Mission.Status.unassigned, Mission.Status.tentative].includes(mission.status);
```

Makes it more clear we're only using once

src/app/component/MissionComponent/MissionDetailsCard/MissionDetailsCard.jsx
```diff
@@ -66,6 +70,66 @@ const Row = ({ children, Icon, label }) => {
...
+`;
+const TextRowEditable = ({ disabled, label, onClick, text }) => {
+  const [value, setValue] = useState(text);
+  const [isEdit, setIsEdit] = useState(false);
+
+  function handleChange(event) {
+    setValue(event.target.value);
+  }
+  //disabled have higher priority
+  let isEditable = isEdit;
```
qdozaq, line 87: Could we do something similar to how you handled line 95?

It's just odd to reassign variables like this especially when they're already tied to a state

src/app/component/MissionComponent/MissionDetailsCard/MissionDetailsCard.jsx
mat10tng: yeah, you are right! It should look like that

src/app/component/MissionComponent/MissionDetailsCard/MissionDetailsCard.jsx
mat10tng: yeah, you are right

src/app/component/MissionComponent/MissionDetailsCard/MissionDetailsCard.jsx
mat10tng: resolved

mat10tng:
You guys can merge in accordingly

PR #636, opened 2020-05-28
Feat/427 add feedback

qdozaq:
feedback success
feedback bad


src/app/page/Recipient/Dashboard.tsx
```diff
@@ -42,27 +43,49 @@ const completedStatus = [MissionStatus.delivered, MissionStatus.succeeded];
...
   const [missions, setMissions] = useState<{
     submitted: MissionInterface[];
     completed: MissionInterface[];
   }>({ submitted: [], completed: [] });
 
+  // @ts-ignore no types for state
+  const auth = useSelector((state) => state.firebase.auth);
+  const org = useOrganization();
+
+  useFirestoreConnect(() => {
```
qdozaq, line 55: swapping out the `getAllRequestedMissions` call for this so we can update on feedback changes

src/app/page/Recipient/MissionFeedback.jsx
```diff
@@ -0,0 +1,129 @@
...
+    top: "0",
+    right: "0",
+    "& svg": {
+      width: "2.5rem",
+      height: "2.5rem",
+    },
+  },
+}));
+
+const MissionFeedback = ({ deliveryImage, missionUid, setShow, show }) => {
```
qdozaq, line 45: This component was moved from `src/app/page/MissionFeedback/MissionFeedback.jsx`

Main changes here are updating the way we submit and wrapping everything in the modal and slide components

src/app/page/Recipient/MissionFeedback.jsx
```diff
@@ -0,0 +1,129 @@
...
+    width: "100%",
+    maxWidth: theme.breakpoints.width("sm"),
+    position: "relative",
+    height: "100%",
+  },
+  modal: {
+    display: "flex",
+    alignItems: "center",
+    justifyContent: "center",
+    height: "100vh",
```
amylo, line 32: Might want to verify on an actual mobile device, that100vh looks ok. I was styling our navigation hamburger menu and had unintentional consequences in a mobile phone browser due to the address bar actually consuming vertical space on the screen... so the content ended up being cut off.
So I ended up using `100%` instead.

src/app/page/Recipient/MissionFeedback.jsx
```diff
@@ -0,0 +1,129 @@
...
+      Content = <FeedbackWithSuccess {...contentProps} handleSubmit={submitFeedback} />;
+      break;
+    case Step.ACK:
+      Content = <Acknowledgement {...contentProps} />;
+      break;
+    default:
+      Content = <DeliveryConfirmation {...contentProps} />;
+  }
+  return (
+    <Modal className={classes.modal} open={show} BackdropComponent={Backdrop}>
```
amylo, line 116: Do we want to materialize this modal experience (with animations, icons, etc ) as its own reusable component so that we can have a consistent modal behavior across the app?

src/app/page/Recipient/MissionFeedback.jsx
qdozaq: mmmm good point. ugh I hate the mobile browser experience

src/app/page/Recipient/MissionFeedback.jsx
qdozaq: Yah I'll go ahead and break it out 👍

PR #646, opened 2020-07-28
Disable failing seed db action

jwu910:
Current seed db action has been failing for months. Until a need is established and a fix is created, can we delete the seed action for now?


utunga:
I know that it shows an error in the log but i wonder is it actually failing - i only ask because i see that the data in the staging site was actually being reset even though it had failing logs _ at least that was the case a while ago... ?

jwu910:
The action is failing, i've been getting notifications every 2 hours for the last 2 months =/ lol. I've just tried to ignore them haha. Maybe someone can look at it and add it back in

